PR #87, opened 2022-04-07
Fix EVM block hash opcode range checking

PlasmaPower:


hkalodner (approved):
LGTM

18dew:
Will kaccak hash change because of this change

PR #93, opened 2022-04-19
Support `eth_feeHistory`

RachelFranks:
Supports `eth_feeHistory` by providing an output with similar semantic meaning & shape to the vanilla rpc call

### DIFF
--- a/arbitrum/apibackend.go
+++ b/arbitrum/apibackend.go
@@ -129,6 +129,9 @@ func (a *APIBackend) FeeHistory(
 	for i := range rewards {
 		rewards[i] = zeros
 	}
+	if len(rewardPercentiles) == 0 {
+		rewards = nil
+	}
 
 	gasUsed := make([]float64, blocks)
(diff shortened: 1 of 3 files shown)


arbitrum/apibackend.go
```diff
@@ -84,8 +85,95 @@ func (a *APIBackend) SuggestGasTipCap(ctx context.Context) (*big.Int, error) {
...
+	if blocks < 1 {
+		// returning with no data and no error means there are no retrievable blocks
+		return common.Big0, nil, nil, nil, nil
+	}
+
+	// don't attempt to include blocks before genesis
+	if rpc.BlockNumber(blocks) > newestBlock {
+		blocks = int(newestBlock + 1)
+	}
+	oldestBlock := int(newestBlock) + 1 - blocks
```
hkalodner, line 121: For this I think we'll need to be careful to have the max depth be the nitro genesis block if we're starting from a state port

hkalodner (changes requested):

hkalodner (approved):
LGTM

PR #96, opened 2022-04-27
Fix effective gas price in arbitrum

hkalodner:
Set effective gas price equal to the base fee in the rpc and graphql interfaces


joshuacolvin0 (approved):
LGTM

PR #97, opened 2022-05-05
Update ArbOS version to 4

edfelten:


params/config_arbitrum.go
```diff
@@ -54,7 +54,7 @@ func ArbitrumTestnetParams() ArbitrumChainParams {
 		EnableArbOS:               true,
 		AllowDebugPrecompiles:     false,
 		DataAvailabilityCommittee: false,
-		InitialArbOSVersion:       1,
+		InitialArbOSVersion:       4,
```
PlasmaPower, line 57: We only want to do this for the dev test and dev test das chains (the ones we use for automated tests). This would change the history of the testnet to start as version 4, whereas we want it to start at version 1 and be upgraded to version 4.

PlasmaPower (changes requested):

params/config_arbitrum.go
RachelFranks: Fixed

RachelFranks (approved):
LGTM

PR #100, opened 2022-05-24
L1RequestId missing from copy

yahgwai:
This meant we were losing the L1RequestId which, apart from other thing, was affecting the tx hash.


edfelten (approved):
Good catch.

PR #103, opened 2022-06-01
support arbitrum-legacy state import

RachelFranks:
Mostly Tsahi's work, but I made it a PR. This PR updates geth to rely on the `genesisBlockNumber` chain param rather than global state


tsahee:
created a different PR

PR #105, opened 2022-06-02
Native and Js tracing impls for `CaptureArbitrumTransfer`

RachelFranks:
Implements `CaptureArbitrumTransfer` for the js tracing system, exposing additional info about transfers before and after evm execution for native and js tracers


PlasmaPower (approved):
LGTM

PR #106, opened 2022-06-09
Remove tips

RachelFranks:
Removes tips for the devnet reinit now that they've been unused since ArbOS version 2


PlasmaPower (approved):
LGTM

PR #108, opened 2022-06-10
Fix fee history with exponential pricing and set ArbOS versions back to 1

PlasmaPower:


core/arbitrum_hooks.go
```diff
@@ -38,8 +38,8 @@ var InterceptRPCMessage func(
 	backend NodeInterfaceBackendAPI,
 ) (types.Message, *ExecutionResult, error)
 
-// Gets ArbOS's approximation of how quickly compute gas is being burnt relative to the speed limit.
-var GetArbOSComputeRate func(statedb *state.StateDB) (float64, error)
+// Gets ArbOS's maximum gas per second
```
RachelFranks, line 41: maximum _intended_ gas per second

arbitrum/apibackend.go
```diff
@@ -128,41 +128,59 @@ func (a *APIBackend) FeeHistory(
...
+				blockGasUsed += receipt.GasUsed - receipt.GasUsedForL1
+			}
+		}
+
+		timeSinceLastBlock := header.Time - prevTimestamp
+		prevTimestamp = header.Time
+
+		// In vanilla geth, this RPC returns the gasUsed ratio so a client can infer how the basefee will change
+		// To emulate this, we translate the compute rate into something like that, centered at an analogous 0.5
+		fullnessAnalogue := float64(blockGasUsed) / float64(speedLimit) / float64(timeSinceLastBlock) / 2.0
```
RachelFranks, line 178: `timeSinceLastBlock` could be `0`

RachelFranks (changes requested):
The consensus changes look great, but there's some minor changes needed for the RPC

arbitrum/apibackend.go
PlasmaPower: I've attempted to resolve this by keeping a running total of the current timestamp, instead of the current block. It's not perfect but I think it's a good method.

PR #110, opened 2022-06-15
Genesisblock chainconfig

tsahee:


params/config_arbitrum.go
```diff
@@ -38,6 +39,28 @@ func (c *ChainConfig) DebugMode() bool {
...
+func (c *ChainConfig) checkArbitrumCompatible(newcfg *ChainConfig, head *big.Int) *ConfigCompatError {
+	boolToBig := func(b bool) *big.Int {
+		if b {
+			return common.Big1
+		}
+		return common.Big0
+	}
+
+	if c.IsArbitrum() != newcfg.IsArbitrum() {
+		return newCompatError("isArbitrum", boolToBig(c.IsArbitrum()), boolToBig(newcfg.IsArbitrum()))
```
PlasmaPower, line 51: I believe both arguments in the `newCompatError` should be `common.Big0`, as if a chain is Arbitrum is determined at the genesis. The arguments to `newCompatError` do not need to necessarily differ -- see the chain ID check for instance.

PlasmaPower (commented):
A couple minor things I'd recommend changing

PlasmaPower (approved):
LGTM. I'll make a follow up PR to address my comments but they're very minor and I'd like to get state transition changes merged.

PR #111, opened 2022-06-15
Improve Arbitrum config compat errors

PlasmaPower:
- Moves the `checkArbitrumCompatible` to earlier in the function to avoid merge conflicts
- Returns block 0 as the divergence when one chain has Arbitrum enabled and the other doesn't


tsahee (approved):
LGTM

PR #114, opened 2022-06-16
support Arbitrum legacy transactions and receipts

tsahee:


core/types/transaction_marshalling.go
```diff
@@ -358,41 +358,51 @@ func (t *Transaction) UnmarshalJSON(input []byte) error {
...
-			Hash:     dec.Hash,
-			Data:     *dec.Data,
-			Nonce:    uint64(*dec.Nonce),
-			To:       dec.To,
-			Value:    (*big.Int)(dec.Value),
-			V:        (*big.Int)(dec.V),
-			R:        (*big.Int)(dec.R),
-			S:        (*big.Int)(dec.S),
+			LegacyTx:     itx,
+			HashOverride: dec.Hash,
```
tsahee, line 405: should also initialize:
	EffectiveGasPrice uint64
	L1BlockNumber     uint64

core/types/transaction_marshalling.go
```diff
@@ -356,41 +360,59 @@ func (t *Transaction) UnmarshalJSON(input []byte) error {
...
-			GasPrice: (*big.Int)(dec.GasPrice),
-			Hash:     dec.Hash,
-			Data:     *dec.Data,
-			Nonce:    uint64(*dec.Nonce),
-			To:       dec.To,
-			Value:    (*big.Int)(dec.Value),
-			V:        (*big.Int)(dec.V),
-			R:        (*big.Int)(dec.R),
-			S:        (*big.Int)(dec.S),
+			LegacyTx:          itx,
```
PlasmaPower, line 412: We should add a test case on the nitro side that we can fetch a block via an RPC eth client with legacy transactions, which validates both serialization and deserialization

params/config_arbitrum.go
```diff
@@ -282,11 +286,37 @@ func ArbitrumAnytrustGoerliTestnetChainConfig() *ChainConfig {
...
+		EIP155Block:         big.NewInt(0),
+		EIP158Block:         big.NewInt(0),
+		ByzantiumBlock:      big.NewInt(0),
+		ConstantinopleBlock: big.NewInt(0),
+		PetersburgBlock:     big.NewInt(0),
+		IstanbulBlock:       big.NewInt(0),
+		MuirGlacierBlock:    big.NewInt(0),
+		BerlinBlock:         big.NewInt(0),
+		LondonBlock:         big.NewInt(0),
+		ArbitrumChainParams: ArbitrumDevTestParams(),
```
PlasmaPower, line 306: Let's use a different `ArbitrumChainParams` for this, since I don't want to accidentally enable `AllowDebugPrecompiles` for rinkeby

PlasmaPower (commented):
Generally LGTM

params/config_arbitrum.go
tsahee: done. LMK if the new params seem o.k.

core/types/transaction_marshalling.go
tsahee: created nito issue https://github.com/OffchainLabs/nitro/issues/830

params/config_arbitrum.go
```diff
@@ -96,6 +96,17 @@ func ArbitrumRollupGoerliTestnetParams() ArbitrumChainParams {
 	}
 }
 
+func ArbitrumRinkebyTestParams() ArbitrumChainParams {
+	return ArbitrumChainParams{
+		EnableArbOS:               true,
+		AllowDebugPrecompiles:     false,
+		DataAvailabilityCommittee: false,
+		// Not used - has init data
+		InitialArbOSVersion: 1,
```
PlasmaPower, line 105: Isn't InitialArbOSVersion still used with init data? I'm not sure how else we'd determine the initial ArbOS version. I think we should set it to 3.

params/config_arbitrum.go
```diff
@@ -92,6 +96,17 @@ func ArbitrumRollupGoerliTestnetParams() ArbitrumChainParams {
...
 }
 
+func ArbitrumRinkebyTestParams() ArbitrumChainParams {
+	return ArbitrumChainParams{
+		EnableArbOS:               true,
+		AllowDebugPrecompiles:     false,
+		DataAvailabilityCommittee: false,
+		InitialArbOSVersion:       3,
+		// Not used - has init data
+		InitialChainOwner: common.Address{},
```
PlasmaPower, line 106: Looking over the code again, I'm not sure we migrate chain owners, so this might be needed. I'd like to get this PR merged though, so I'll just make a follow-up PR to correct the comment.

PlasmaPower (approved):
LGTM

params/config_arbitrum.go
PlasmaPower: https://github.com/OffchainLabs/go-ethereum/pull/132

PR #116, opened 2022-06-17
Add From field to Arbitrum Deposit Tx

PlasmaPower:
I merged #112 a bit too early. The corresponding nitro change is merged into https://github.com/OffchainLabs/nitro/pull/692/files but not https://github.com/OffchainLabs/nitro/pull/585 . This restores the separation.


RachelFranks (approved):
LGTM

PR #120, opened 2022-06-24
Fix copying the L1BaseFee in a retryable submission

PlasmaPower:
This shouldn't matter in practice, as I highly doubt we mutate the L1BaseFee of a retryable, but it's good to fix this for future reference


RachelFranks (approved):
LGTM

PR #122, opened 2022-06-27
`RetryValue` field

RachelFranks:
Renames `ArbitrumSubmitRetryableTx`'s `Value` to `RetryValue` and reports `0` for calls to `Value()`

This does not change consensus


PlasmaPower (approved):
LGTM

PR #128, opened 2022-07-14
Add hook for gasprice opcode

RachelFranks:
Adds a `GasPriceOp` hook so that ArbOS can cause this opcode to not include the tip in version 3


PlasmaPower (approved):
LGTM

PR #129, opened 2022-07-15
Use ArbOS version 3 in testing

edfelten:
This set ArbOS version to 3 in Nitro testing configurations.


PlasmaPower:
I believe this change is already in master

RachelFranks:
Closing since these changes were adopted in master

PR #130, opened 2022-07-18
Rpc fallback

tsahee:


arbitrum/apibackend.go
```diff
@@ -449,3 +502,11 @@ func (a *APIBackend) Engine() consensus.Engine {
 func (b *APIBackend) PendingBlockAndReceipts() (*types.Block, types.Receipts) {
 	return nil, nil
 }
+
+func (b *APIBackend) FallbackClient() types.FallbackClient {
+	if b.fallbackClient == nil && b.fallbackClientUrl != "" {
+		// TODO: reattempt connection?
+		log.Error("fallback client requested, but initial connection failed")
```
PlasmaPower, line 509: We should definitely either fail on boot if the connection to the fallback node fails, or reconnect here. As-is, it seems way too likely that we'd have some intermittent networking issue causing connections to the fallback nodes to fail, but the issue would persist long after we fixed the cause because any nitro nodes that started up during that time period would be unable to fallback.

internal/ethapi/api.go
```diff
@@ -1224,7 +1251,13 @@ func (s *PublicBlockChainAPI) EstimateGas(ctx context.Context, args TransactionA
 	if blockNrOrHash != nil {
 		bNrOrHash = *blockNrOrHash
 	}
-	return DoEstimateGas(ctx, s.b, args, bNrOrHash, s.b.RPCGasCap())
+	res, err := DoEstimateGas(ctx, s.b, args, bNrOrHash, s.b.RPCGasCap())
+	if client := fallbackClientFor(s.b, err); client != nil {
+		var res hexutil.Uint64
+		err := client.CallContext(ctx, &res, "eth_estimateGas", args, blockNrOrHash)
```
PlasmaPower, line 1257: This fails against classic:
```
lee-arch :: ~ » eth 'http://localhost:8549/rpc' estimateGas '{}' pending
{
  "error": {
    "code": -32602,
    "message": "too many arguments, want at most 1"
  }
}
```
That said, it doesn't accept the block number argument, we have no way to serve that estimate gas against old classic blocks anyways. The best we could do is serve it against the last classic block which probably isn't very useful. So I'd say it's fine to keep this as-is, just pointing this out.

PlasmaPower (changes requested):

internal/ethapi/api.go
hkalodner: It'd be easy to add support for the extra param to classic. We probably might as well include that in the next release even though it's hard to imagine a use case for it

PlasmaPower (approved):
LGTM

PR #137, opened 2022-07-29
SyncProgressMap

tsahee:
allows a more flexible eth_sync

### DIFF
--- a/arbitrum/apibackend.go
+++ b/arbitrum/apibackend.go
@@ -9,11 +9,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/eth"
 	"github.com/ethereum/go-ethereum/eth/tracers"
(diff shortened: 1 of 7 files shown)


arbitrum/apibackend.go
```diff
@@ -130,8 +136,20 @@ func (a *APIBackend) GetArbitrumNode() interface{} {
 }
 
 // General Ethereum API
+func (a *APIBackend) SyncProgressMap() map[string]interface{} {
+	return a.sync.SyncProgressMap()
+}
+
 func (a *APIBackend) SyncProgress() ethereum.SyncProgress {
-	panic("not implemented") // TODO: Implement
+	progress := a.sync.SyncProgressMap()
```
PlasmaPower, line 144: The impl looks good, but I'm curious, where is `SyncProcess()` called?

tsahee: from graphql's syncprogress - which requires the ethereum-defined struct

PlasmaPower (approved):
LGTM

PR #139, opened 2022-07-29
set ErrUseFallback code and message directly

tsahee:
this uses the same error message even if not wrapped by a fallback client call (e.g. debug namespace)

### DIFF
--- a/core/types/arb_types.go
+++ b/core/types/arb_types.go
@@ -22,7 +22,7 @@ var fallbackErrorCode = -32001
 func SetFallbackError(msg string, code int) {
 	fallbackErrorMsg = msg
 	fallbackErrorCode = code
-	log.Info("setting fallback", "msg", msg, "code", code)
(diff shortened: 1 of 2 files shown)


core/types/arb_types.go
```diff
@@ -3,17 +3,32 @@ package types
...
 
 	"github.com/ethereum/go-ethereum/common"
 )
 
-var ErrUseFallback = errors.New("missing trie node 0000000000000000000000000000000000000000000000000000000000000000 (path )")
+type fallbackError struct {
+}
+
+var fallbackErrorMsg = "I seem to be missing trie node 0000000000000000000000000000000000000000000000000000000000000000 (path )"
+var fallbackErrorCode = -32001
```
joshuacolvin0, line 20: I think the default should be the same as geth to maximize compatibility, with the option of customizing it

joshuacolvin0 (dismissed):
LGTM

PlasmaPower (approved):
LGTM

PR #143, opened 2022-08-05
add l1blockNumber for classic blocks

tsahee:
This adds l1BlockNumber to classic blocks.
BlockNumber is taken from first transaction. If that doesn't exist - first transaction of next block with one. Up to 5 blocks are scanned (in rinkeby there is at most a 2-block empty streak)

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1333,15 +1333,57 @@ func (s *PublicBlockChainAPI) rpcMarshalHeader(ctx context.Context, header *type
 	return fields
 }
 
+func (s *PublicBlockChainAPI) arbClassicL1BlockNumber(ctx context.Context, block *types.Block) (hexutil.Uint64, error) {
+	chainConfig := s.b.ChainConfig()
+	blockNum := block.Number().Int64()
+	i := 0
+	for {
+		transactions := block.Transactions()
+		if len(transactions) > 0 {
+			legacyTx, ok := transactions[0].GetInner().(*types.ArbitrumLegacyTxData)
+			if !ok {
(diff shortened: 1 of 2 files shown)


internal/ethapi/api.go
```diff
@@ -1333,16 +1333,58 @@ func (s *PublicBlockChainAPI) rpcMarshalHeader(ctx context.Context, header *type
...
+	for {
+		transactions := block.Transactions()
+		if len(transactions) > 0 {
+			legacyTx, ok := transactions[0].GetInner().(*types.ArbitrumLegacyTxData)
+			if !ok {
+				return 0, fmt.Errorf("couldn't read legacy transaction from block %d", blockNum)
+			}
+			return hexutil.Uint64(legacyTx.L1BlockNumber), nil
+		}
+		blockNum++
```
PlasmaPower, line 1349: Shouldn't we be subtracting one from the block number here, not adding one?

tsahee: Both should be good enough.
Blocks 0 and 1 don't have transactions in rinkeby. We could decide to return 0 or nothing in that case and subtract one.
Adding one requires solving the other edge case - getting to nitro genesis, though I don't think we should really hit that, as the last block does have a transaction.
Solution for this edge case seemed cleaner so I went with that - but I can change if you think that's better.

joshuacolvin0: I kindof like returning the block with the last executed transaction, so I'm going to side with Lee.  If we return the first block with executed transaction after the current block, it is kindof like predicting the future.  And then a special case where getting to block zero would just return block zero without error sounds reasonable.

PlasmaPower (approved):
LGTM

PR #149, opened 2022-08-19
Better default hooks

rachel-bousfield:
Improves the default values for the intercept hooks


internal/ethapi/api.go
```diff
@@ -731,10 +731,10 @@ func (s *PublicBlockChainAPI) GetHeaderByHash(ctx context.Context, hash common.H
...
 
 // GetBlockByNumber returns the requested canonical block.
-// * When blockNr is -1 the chain head is returned.
-// * When blockNr is -2 the pending chain head is returned.
-// * When fullTx is true all transactions in the block are returned, otherwise
-//   only the transaction hash is returned.
+//   - When blockNr is -1 the chain head is returned.
+//   - When blockNr is -2 the pending chain head is returned.
+//   - When fullTx is true all transactions in the block are returned, otherwise
+//     only the transaction hash is returned.
```
PlasmaPower, line 737: Let's undo this to avoid upstream diff

internal/ethapi/api.go
```diff
@@ -965,7 +965,7 @@ func DoCall(ctx context.Context, b Backend, args TransactionArgs, blockNrOrHash
 	msg.TxRunMode = txRunMode
 
 	// Arbitrum: support NodeInterface.sol by swapping out the message if needed
-	if core.InterceptRPCMessage != nil {
+	{
```
PlasmaPower, line 968: We can just pull the code out of this block now that there's no longer an if statement

PlasmaPower (changes requested):

internal/ethapi/api.go
RachelFranks: Oops, didn't realize my editor did this

internal/ethapi/api.go
RachelFranks: Sure, I was trying to contain the logic in a single scope but it's not necessary

hkalodner (approved):
LGTM

PlasmaPower (approved):
LGTM

PR #151, opened 2022-08-23
Add "finalize" option to `gethclient`

RachelFranks:
Adds the ability for the user to specify "finalize" as they would "pending"


PlasmaPower (approved):
LGTM

PR #155, opened 2022-08-29
support non-clean shutdown

tsahee:
Adds shutdown markers, and adds time on blockchain which invokes triedb cache writes


joshuacolvin0 (approved):
LGTM

core/blockchain.go
```diff
@@ -1339,6 +1339,16 @@ func (bc *BlockChain) WriteBlockAndSetHead(block *types.Block, receipts []*types
...
 }
 
+// WriteBlockAndSetHeadWithTime also counts processTime, which will cause intermittent TrieDirty cache writes
+func (bc *BlockChain) WriteBlockAndSetHeadWithTime(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB, emitHeadEvent bool, processTime time.Duration) (status WriteStatus, err error) {
+	if !bc.chainmu.TryLock() {
+		return NonStatTy, errChainStopped
+	}
+	defer bc.chainmu.Unlock()
+	bc.gcproc += processTime
+	return bc.writeBlockAndSetHead(block, receipts, logs, state, emitHeadEvent)
```
PlasmaPower, line 1349: Let's put this in blockchain_arbitrum.go to prevent merge conflicts

PlasmaPower (dismissed):
LGTM

core/blockchain_arbitrum.go
```diff
@@ -54,3 +67,52 @@ func (bc *BlockChain) ClipToPostNitroGenesis(blockNum rpc.BlockNumber) (rpc.Bloc
 	}
 	return blockNum, currentBlock
 }
+
+// finds the number of blocks that aren't prunable
+func (bc *BlockChain) FindRetentionBound() uint64 {
```
hkalodner, line 72: If `TrieRetention` is 0, the behavior should be the same as standard geth, right? Might be good to have an early exit to make that clear

hkalodner (changes requested):

core/blockchain_arbitrum.go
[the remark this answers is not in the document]
tsahee: this would work better with a little-different search algo.. leaving it to later

hkalodner (approved):
LGTM

PR #156, opened 2022-08-30
commit triedb on cap

tsahee:
cap seems to assume that the bottom layeys triedb is already committed.
Make sure this is true.


tsahee:
not the right way

PR #159, opened 2022-09-02
limit unrolling when searching for a snapshot.

tsahee:
Unrolling to find snapshot could save a lot of time avoiding the need to regenerate it, but it could also force a too-long rewind which is just not worth it. Allow user to set a gas-limit to decide.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -552,7 +552,7 @@ func (bc *BlockChain) SetSafe(block *types.Block) {
 // setHeadBeyondRoot rewinds the local chain to a new head with the extra condition
 // that the rewind must pass the specified state root. The extra condition is
 // ignored if it causes rolling back more than rewindLimit Gas (0 meaning infinte).
-// If the limit was hit it will choose latest full head. This method is meant to be
+// If the limit was hit, rewind to last block with state. This method is meant to be
 // used when rewinding with snapshots enabled to ensure that we go back further than
 // persistent disk layer. Depending on whether the node was fast synced or full, and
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -314,17 +316,20 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
...
 			if err != nil {
 				return nil, err
 			}
 			// Chain rewound, persist old snapshot number to indicate recovery procedure
-			if snapDisk != 0 {
+			if diskRootFound {
 				rawdb.WriteSnapshotRecoveryNumber(bc.db, snapDisk)
+			} else {
+				log.Warn("Snapshot root not found or too far back. Recreating snapshot from scratch.")
+				rawdb.DeleteSnapshotRecoveryNumber(bc.db)
```
PlasmaPower, line 328: Do we also need to delete the snapshot itself here?

core/blockchain.go
```diff
@@ -618,6 +632,14 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, root common.Hash, repair bo
 						log.Debug("Rewound to block with state", "number", newHeadBlock.NumberU64(), "hash", newHeadBlock.Hash())
 						break
 					}
+					if lastFullBlock != 0 && rewindLimit > 0 {
+						gasRolledBack += newHeadBlock.GasUsed()
```
PlasmaPower, line 636: We need to subtract out the gas used for L1 in receipts for arbitrum

PlasmaPower (approved):
LGTM

PR #160, opened 2022-09-05
arb_legacy_tx allow sender override

tsahee:


tsahee:
I'm marking this as draft. It's ready for review - but I specifically want to avoid merging it to master until the arbitrum PR is merged 
https://github.com/OffchainLabs/arbitrum/pull/2495

core/types/arbitrum_legacy_tx.go
```diff
@@ -13,9 +14,25 @@ type ArbitrumLegacyTxData struct {
...
+	EffectiveGasPrice uint64
+	L1BlockNumber     uint64
+}
+
+type arbitrumLegacyTxDataWithSenderRLP struct {
+	LegacyTx
+	HashOverride      common.Hash // Hash cannot be locally computed from other fields
+	EffectiveGasPrice uint64
+	L1BlockNumber     uint64
+	Sender            common.Address // only used in unsigned Txs
```
PlasmaPower, line 32: What I'd recommend doing here instead of having two structs is adding a new field which looks like this:
```go
Sender *common.Address `rlp:"optional"`
```
This should let it deserialize transaction data with or without a sender, without using two different structs.

tsahee: nice! didn't know about that one, Fixed.

PlasmaPower (approved):
LGTM

PR #161, opened 2022-09-09
Lower `arbdebug` defaults

rachel-bousfield:
Decreases the default max amount of historical data returned by the node


PlasmaPower (approved):
LGTM

PR #165, opened 2022-09-14
blockchain: in arbitrum, genesis is not block 0

tsahee:
Specifically, this makes sure setHeadBeyondRoot does not rewind the chain to before nitro's genesis block.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -274,7 +274,11 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 	if err != nil {
 		return nil, err
 	}
-	bc.genesisBlock = bc.GetBlockByNumber(0)
+	if chainConfig.IsArbitrum() {
+		bc.genesisBlock = bc.GetBlockByNumber(chainConfig.ArbitrumChainParams.GenesisBlockNum)
+	} else {
+		bc.genesisBlock = bc.GetBlockByNumber(0)
+	}
 	if bc.genesisBlock == nil {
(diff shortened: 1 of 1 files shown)


core/blockchain.go
```diff
@@ -274,7 +274,11 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 	if err != nil {
 		return nil, err
 	}
-	bc.genesisBlock = bc.GetBlockByNumber(0)
+	if chainConfig.IsArbitrum() {
+		bc.genesisBlock = bc.GetBlockByNumber(chainConfig.ArbitrumChainParams.GenesisBlockNum)
+	} else {
+		bc.genesisBlock = bc.GetBlockByNumber(0)
```
PlasmaPower, line 280: I'm a bit concerned about this change as it impacts everyone querying for the genesis block (which they might assume is block 0), but my preliminary search of the codebase didn't turn up anything dangerous

PlasmaPower (approved):
LGTM

core/blockchain.go
tsahee: The only place I saw so far that access the genesis block (via Genesis() function), not during a test is forkid, accoring to https://eips.ethereum.org/EIPS/eip-2124 - I suppose there could be arguments both ways, and either block0 or nitro genesis will ultimately be fine.
In tests (e.g. graphql) as well as here I saw genesis used as the first block from which others are built - in which case we definitely want the nitro genesis.

PR #166, opened 2022-09-15
Make newRevertError have consistent prefix

hkalodner:
This makes it so the `newRevertError` function always returns an error message starting with "execution reverted"


joshuacolvin0 (approved):
LGTM

PR #169, opened 2022-10-03
Recording fixes

tsahee:
recording fixes.
If replay uses state.NewDeterministic, PrepareRecording can be called with validateDeleted = false. Otherwise true is required (will collect more preimages)

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -322,11 +322,7 @@ func (s *stateObject) updateTrie(db Database) Trie {
 	hasher := s.db.hasher
 
 	usedStorage := make([][]byte, 0, len(s.pendingStorage))
-	type hashAndBig struct {
-		hash common.Hash
-		big  *big.Int
-	}
-	keysToDelete := make([]hashAndBig, 0, len(s.pendingStorage))
+	keysToDelete := make([]common.Hash, 0, len(s.pendingStorage))
 	for key, value := range s.pendingStorage {
 		// Skip noop changes, persist actual changes
 		if value == s.originStorage[key] {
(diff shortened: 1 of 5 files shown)


core/state/state_object.go
```diff
@@ -351,6 +361,13 @@ func (s *stateObject) updateTrie(db Database) Trie {
 		}
 		usedStorage = append(usedStorage, common.CopyBytes(key[:])) // Copy needed for closure
 	}
+	if len(keysToDelete) > 0 {
+		sort.Slice(keysToDelete, func(i, j int) bool { return keysToDelete[i].big.Cmp(keysToDelete[j].big) < 0 })
```
PlasmaPower, line 365: I think we can directly compare `common.Hash`s as they're a `[32]byte` right? It's worth a try to simplify this I think.

core/state/statedb.go
```diff
@@ -890,9 +901,26 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
...
+	if s.deterministic {
+		type addressAndBig struct {
+			address common.Address
+			big     *big.Int
+		}
+		addressesToUpdate := make([]addressAndBig, 0, len(s.stateObjectsPending))
+		for addr := range s.stateObjectsPending {
+			addressesToUpdate = append(addressesToUpdate, addressAndBig{addr, addr.Hash().Big()})
+		}
+		sort.Slice(addressesToUpdate, func(i, j int) bool { return addressesToUpdate[i].big.Cmp(addressesToUpdate[j].big) < 0 })
```
PlasmaPower, line 913: Same thing here, I think we can directly compare addresses

PlasmaPower (changes requested):

PlasmaPower (approved):
LGTM

PlasmaPower (approved):
LGTM

PR #172, opened 2022-10-20
Tx Tips in L2

rachel-bousfield:
Passes the info needed to compute tips to the tx processor

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -345,6 +345,7 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	}
 	st.gas -= gas
 
+	tipAmount := big.NewInt(0)
 	tipReceipient, err := st.evm.ProcessingHook.GasChargingHook(&st.gas)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


core/state_transition.go
```diff
@@ -388,9 +389,9 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	} else {
 		fee := new(big.Int).SetUint64(st.gasUsed())
 		fee.Mul(fee, effectiveTip)
-		st.state.AddBalance(st.evm.Context.Coinbase, fee)
+		st.state.AddBalance(tipReceipient, fee)
 
-		// Arbitrum: record the tip if nonzero (this should never happen in L2)
+		// Arbitrum: record the tip
 		if st.evm.Config.Debug {
 			st.evm.Config.Tracer.CaptureArbitrumTransfer(st.evm, nil, &st.evm.Context.Coinbase, fee, false, "tip")
```
PlasmaPower, line 396: We should probably always record this if `!DropTip()` for clarity (we also record other zero transfers)

PlasmaPower (commented):
Generally LGTM, one minor tracing comment

core/state_transition.go
rachel-bousfield: Moved below so we can do this :)

PlasmaPower (approved):
LGTM

PR #177, opened 2022-11-02
recording database updates

tsahee:
This converges support needed for validation without an archive node.

### DIFF
--- a/arbitrum/recordingdb.go
+++ b/arbitrum/recordingdb.go
@@ -310,6 +310,7 @@ func (r *RecordingDatabase) GetOrRecreateState(ctx context.Context, header *type
 		if err != nil {
 			return nil, fmt.Errorf("failed commiting state for block %d : %w", blockToRecreate, err)
 		}
+		r.dereferenceRoot(lastRoot)
 		lastRoot = block.Root()
 		if blockToRecreate >= returnedBlockNumber {
 			if block.Hash() != header.Hash() {
@@ -322,4 +323,8 @@ func (r *RecordingDatabase) GetOrRecreateState(ctx context.Context, header *type
 		blockToRecreate++
 	}
 	return nil, ctx.Err()
+}
+
+func (r *RecordingDatabase) ReferenceCount() int64 {
+	return r.references
 }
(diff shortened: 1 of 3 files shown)


arbitrum/recordingdb.go
```diff
@@ -167,29 +244,94 @@ func PrepareRecording(blockchain *core.BlockChain, lastBlockHeader *types.Header
...
+		currentHeader = r.bc.GetHeader(currentHeader.ParentHash, currentHeader.Number.Uint64()-1)
+		if currentHeader == nil {
+			return nil, fmt.Errorf("chain doesn't contain parent of block %d hash %v", currentHeader.Number, currentHeader.Hash())
+		}
+		stateDb, err = r.StateFor(currentHeader)
+		if err == nil {
+			lastRoot = currentHeader.Root
+			defer func() {
+				if (lastRoot != common.Hash{}) {
+					r.dereferenceRoot(lastRoot)
```
PlasmaPower, line 297: Let's move this defer out of the for loop to make it a bit more visible and clarify when it'll run

arbitrum/recordingdb.go
```diff
@@ -167,29 +244,94 @@ func PrepareRecording(blockchain *core.BlockChain, lastBlockHeader *types.Header
...
+				}
+			}()
+			break
+		}
+	}
+	blockToRecreate := currentHeader.Number.Uint64() + 1
+	for ctx.Err() == nil {
+		block := r.bc.GetBlockByNumber(blockToRecreate)
+		if block == nil {
+			return nil, fmt.Errorf("block not found while recreating: %d", blockToRecreate)
```
PlasmaPower, line 307: We could check here if `block.ParentHash()` matches what we'd expect to provide early protection against reorgs, though I suppose checking that the state root matches is also good enough

PlasmaPower (commented):
I have a couple minor comments but I love the approach!

arbitrum/recordingdb.go
tsahee: done. Worth noting, though, that most reorgs will only be found after we reach the block number we're originally looking for and comparing it with the param header

PlasmaPower (approved):
LGTM

PR #178, opened 2022-11-06
Fix tracking of gasRolledBack

PlasmaPower:
This moves the tracking above the `continue` in case of missing state

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -597,6 +597,24 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, root common.Hash, repair bo
 				gasRolledBack := uint64(0)
 
 				for {
+					if rewindLimit > 0 {
+						// Arbitrum: track the amount of gas rolled back and stop the rollback early if necessary
+						gasUsedInBlock := newHeadBlock.GasUsed()
+						if bc.chainConfig.IsArbitrum() {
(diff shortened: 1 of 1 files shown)


tsahee:
I intentionally only counted rolled-back gas from the first block-with-state, because rolling up to that block is not avoidable anyway, so you just configure how much extra rollback do you allow for snapshot.
E.g. if snapshot is one block behind the first block with full state, but this block is very far back - this patch will discard the snapshot, which will cost much more time than rolling one more block back and using it for the rest of the state.
If you want to limit rollback further - I'd argue lowering that limit significantly is better than moving it to include blocks without state.

PlasmaPower:
I see, but wouldn't it also skip blocks without state after the snapshot?

core/blockchain.go
```diff
@@ -597,6 +597,24 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, root common.Hash, repair bo
 				gasRolledBack := uint64(0)
 
 				for {
+					if rewindLimit > 0 {
```
tsahee, line 600: && lastFullblock != 0

tsahee:
ooh you're right! missed that, nice catch!
just a small comment in the same spirit, then :)

core/blockchain.go
PlasmaPower: Fixed :) thanks!

tsahee (approved):
LGTM

PR #181, opened 2022-11-30
Sliding window

joshuacolvin0:
The current go-ethereum histogram metrics object is set to reset whenever
the metrics endpoint is scraped.  The reset is done to prevent the problem of
histograms not getting updated for commands that are rarely called.  However,
this breaks many Prometheus assumptions and makes computing correct metrics
difficult and unreliable.  The go-ethereum metrics appears to be a fork of
rcrowley/go-metrics which in turn is a port of the Java library
https://github.com/dropwizard/metrics.  Looking at
https://metrics.dropwizard.io/4.2.0/manual/core.html#exponentially-decaying-reservoirs
there is an implementation of `SlidingTimeWindowArrayReservoir`:
```
SlidingTimeWindowArrayReservoir is comparable with
ExponentiallyDecayingReservoir in terms GC overhead and performance. As for
required memory, SlidingTimeWindowArrayReservoir takes ~128 bits per stored
measurement and you can simply calculate required amount of heap.
Example: 10K measurements / sec with reservoir storing time of 1 minute will
take 10000 * 60 * 128 / 8 = 9600000 bytes ~ 9 megabytes
```

Here is more information on the sampling error introduced by Exponential Decay
sampling (unrelated to prometheus issues caused by resetting on every scrape):
https://medium.com/expedia-group-tech/your-latency-metrics-could-be-misleading-you-how-hdrhistogram-can-help-9d545b598374

### DIFF
--- a/eth/protocols/eth/handler.go
+++ b/eth/protocols/eth/handler.go
@@ -219,9 +219,7 @@ func handleMessage(backend Backend, peer *Peer) error {
 		h := fmt.Sprintf("%s/%s/%d/%#02x", p2p.HandleHistName, ProtocolName, peer.Version(), msg.Code)
 		defer func(start time.Time) {
 			sampler := func() metrics.Sample {
-				return metrics.ResettingSample(
-					metrics.NewExpDecaySample(1028, 0.015),
-				)
+				return metrics.NewBoundedHistogramSample()
 			}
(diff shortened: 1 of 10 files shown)


PlasmaPower (approved):
LGTM

PlasmaPower:
I've also verified that the tests pass now that we're on go 1.18 and everything is compiling

PR #184, opened 2022-12-01
Retryable RetryTo field should not be required in JSON unmarshalling

PlasmaPower:
This field is nil in the case of a retryable that functions as a contract creation


rachel-bousfield (approved):
LGTM

PR #186, opened 2022-12-07
Allow pruning to multiple targets, including without snapshots

PlasmaPower:


core/state/pruner/pruner.go
```diff
@@ -261,6 +261,8 @@ func (p *Pruner) Prune(root common.Hash) error {
 		}
 		// Use the bottom-most diff layer as the target
 		root = layers[len(layers)-1].Root()
+	} else if p.snaptree.Snapshot(root) == nil {
+		p.snaptree.Rebuild(root, false)
```
tsahee, line 265: This seems like it will work.. however, I think that a better alternative would be to skip creating the snaptree in NewPruner and passing the right root to snapshot.New which wouldn't require any change in snapshot.

For nitro - I think we'll need to support multiple roots (so all will be cleaned unless in any of these roots), which should probably be done with multiple snaptrees, each with a different root, each updating the same stateBloom.

At least in a validators we'd like to keep state for: something recent + latest validated + latest accepted onchain.

core/state/snapshot/snapshot.go
```diff
@@ -689,7 +695,14 @@ func (t *Tree) Journal(root common.Hash) (common.Hash, error) {
 // Rebuild wipes all available snapshot data from the persistent database and
 // discard all caches and diff layers. Afterwards, it starts a new snapshot
 // generator with the given root hash.
-func (t *Tree) Rebuild(root common.Hash) {
+func (t *Tree) Rebuild(root common.Hash, async bool) {
+	var genPending chan struct{}
+	if !async {
```
tsahee, line 700: 1. Why use defer and not just check async at the end of function?
(if using defer, I'd at least check that genPending is not nil)

2. why not use waitBuild for waiting?

tsahee (commented):
some notes & thoughts

core/state/snapshot/snapshot.go
PlasmaPower: I used a defer because I wanted it to happen after the mutex was unlocked. I'll switch it to waitBuild though.

PlasmaPower: I've removed the need for this change

core/state/pruner/pruner.go
PlasmaPower: I've added support for multiple roots, but I found that having multiple snaptrees was way too expensive, and would completely destroy any existing snapshots. Instead, I've added support for generating the pruning bloom filters without a snapshot :)

core/state/pruner/pruner.go
```diff
@@ -66,9 +66,9 @@ var (
 // Pruner is an offline tool to prune the stale state with the
 // help of the snapshot. The workflow of pruner is very simple:
 //
-// - iterate the snapshot, reconstruct the relevant state
-// - iterate the database, delete all other state entries which
-//   don't belong to the target state and the genesis state
+//   - iterate the snapshot, reconstruct the relevant state
+//   - iterate the database, delete all other state entries which
+//     don't belong to the target state and the genesis state
```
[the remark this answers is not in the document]
PlasmaPower: This is just gofmt which happens every time I save the file. I'd be happy to change it back, but I want to make sure I won't need to make any future changes first, so I'll do it after you approve. Though at this point I've already changed a large percentage of the file.

core/state/pruner/pruner.go
```diff
@@ -228,10 +233,130 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
...
+			return common.Hash{}, errors.New("got zero key for trie db node")
+		}
+		return key, nil
+	} else {
+		return it.Hash(), nil
+	}
+}
+
+// We assume output does not need the value, only the key
+func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output ethdb.KeyValueWriter) error {
```
tsahee, line 253: taking a stateBloo instead of KeyValueWriter would be better for self-documentation

core/state/pruner/pruner.go
```diff
@@ -228,10 +233,130 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
 	return nil
 }
 
+func nodeIteratorKey(it trie.NodeIterator) (common.Hash, error) {
+	if it.Leaf() {
```
tsahee, line 237: Looking through extractGenesis it doesn't have this part - only it.Hash()
Also - if I'm reading this correctly, LeafKey would be the path to the data, and I don't remember that it should appear as a key in the database (?)

tsahee (commented):
I'm mostly comparing dumpTrieDescendant to extractGenesis.
If there is a good reason for the diff - LGTM. Otherwise, code-sharing the two might have nice advantages.

core/state/pruner/pruner.go
PlasmaPower: extractGenesis is right here :) I'm updating my function and having extractGenesis just call mine since mine is parallel and has ETA tracking

tsahee (approved):
LGTM + small suggestion (can be done separately or skipped)

core/state/pruner/pruner.go
```diff
@@ -410,86 +589,29 @@ func extractGenesis(db ethdb.Database, stateBloom *stateBloom) error {
...
-				stateBloom.Put(acc.CodeHash, nil)
-			}
-		}
-	}
-	return accIter.Error()
-}
-
-func bloomFilterName(datadir string, hash common.Hash) string {
-	return filepath.Join(datadir, fmt.Sprintf("%s.%s.%s", stateBloomFilePrefix, hash.Hex(), stateBloomFileSuffix))
+	return dumpRawTrieDescendants(db, genesis.Root(), stateBloom)
```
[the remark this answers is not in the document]
PlasmaPower: Unfortunately we need the genesis block number separately on the nitro side to compute the latest finalized L2 block from the latest finalized message number. Getting the block number is a bit trickier here without the ReadChainConfig helper (it means we'd need to first read block 0, and then read the actual genesis block), so I've kept it on the nitro side for now.

PR #189, opened 2022-12-28
Store bloom filter roots in the file content

PlasmaPower:
Points into https://github.com/OffchainLabs/go-ethereum/pull/186

Fixes the filename being too long for ext4 with 4 roots. An alternative solution would be to base64 the roots in the filename.


PlasmaPower:
I've confirmed that this works on the latest commit so I'm going to merge it into the base PR :)

PR #190, opened 2023-01-04
SlidingWindow metric needs to trim anytime measurements queried

joshuacolvin0:
Need to trim stale data before accessing any SlidingTimeWindow measurements


Tristan-Wilson (approved):
Right, because time may have passed since the last sample was added, which is where `trim` was occurring before. LGTM

PR #193, opened 2023-01-18
Fix log when missing parent while recreating validation state

PlasmaPower:
`currentHeader` is nil in this if statement so the previous log line would panic


magicxyyz (approved):
LGTM

PR #194, opened 2023-01-18
Add a mutex around recreating state in the RecordingDatabase

PlasmaPower:
This isn't necessary for correctness but prevents a ton of duplicate work


PlasmaPower:
Replaced by https://github.com/OffchainLabs/nitro/pull/1443

PR #195, opened 2023-01-19
Mark startup later for the unclean shutdown tracker

PlasmaPower:
If the backend is created but not started, that shouldn't count as an unclean shutdown.


joshuacolvin0 (approved):
LGTM

PR #196, opened 2023-02-03
Support eth_sendRawTransactionConditional

magicxyyz:
adds support for eth_sendRawTransactionConditional

### DIFF
--- a/arbitrum_types/txoptions.go
+++ b/arbitrum_types/txoptions.go
@@ -68,7 +68,6 @@ func (r RootHashOrSlots) MarshalJSON() ([]byte, error) {
 	return json.Marshal(r.SlotValue)
 }
 
-// TODO rename to just "Options" ?
 type ConditionalOptions struct {
 	KnownAccounts  map[common.Address]RootHashOrSlots `json:"knownAccounts"`
 	BlockNumberMin *hexutil.Uint64                    `json:"blockNumberMin,omitempty"`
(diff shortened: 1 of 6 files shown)


arbitrum_types/txoptions.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return o.CheckOnlyStorage(statedb)
+}
+
+func (o *ConditionalOptions) PreCheck(l1BlockNumberLowerBound uint64, statedb *state.StateDB) error {
+	if o.BlockNumberMax != nil && l1BlockNumberLowerBound > uint64(*o.BlockNumberMax) {
+		return NewRejectedError("BlockNumberMax condition not met")
+	}
+	// TODO can we get a timestamp LowerBound?
+	//if o.TimestampMin != nil && l1TimestampLowerBound > uint64(*o.TimestampMax) {
```
PlasmaPower, line 84: The last block's timestamp is a lower bound as timestamp can never decrease.

PlasmaPower: this condition should also be `if o.TimestampMax != nil`

magicxyyz: do you mean L2 block timestamp? or L1 block?

L2 block timestamp is easily available in pre-checker which uses the `PreCheck` method, but I am not sure if L2 timestamp is good enough. My concern is if L1 block timestamp can be older then L2 block timestamp? If so, then L2 block timestamp isn't true lower bound of L1 block timestamp.

If you mean current L1 block timestamp, then I think that getting this value in pre-checker would require some additional operations. Maybe it could be tracked simillar to L1 block number

PlasmaPower: L2 block timestamp, as that's what's visible in the EVM

PlasmaPower (commented):
LGTM but I have some comments on the nitro side

PlasmaPower (approved):
LGTM

PR #203, opened 2023-02-22
Accurate ArbOS storage pricing

rachel-bousfield:
Computes the precise costs of state access as of the Merge so that we can cut fees in ArbOS


rachel-bousfield:
These PRs were intended to assess the complexity of introducing this feature. Given that it turns out to be rather involved and therefore high risk, it doesn't make sense to include these changes unless sufficient demand exists. Will reopen should that change.

PR #204, opened 2023-02-27
Tipping tx type

magicxyyz:
Adds support for new transaction type that wraps DynamicFeeTx.


core/types/transaction_marshalling.go
```diff
@@ -192,6 +193,29 @@ func (t *Transaction) MarshalJSON() ([]byte, error) {
 		data := tx.data()
 		enc.Data = (*hexutil.Bytes)(&data)
 		enc.To = t.To()
+	case *ArbitrumSubtypedTx:
+		subtype := uint64(tx.TxSubtype())
+		enc.Subtype = (*hexutil.Uint64)(&subtype)
+		switch subtype {
+		case ArbitrumTippingTxSubtype:
```
magicxyyz, line 200: alternatively we could call here a method of `ArbitrumSubtypedTx` (something like `tx.(*ArbitrumSubtypedTx).EncodeJSON(&enc)` ) and have the specific tx subtypes implement the encoding. The similar thing could be done for RLP encoding. 
This could be nice to have almost everything that's needed to implement new tx subtype in one place, but also it would take it out of context, so initially I followed geths file split.

PR #210, opened 2023-03-29
Don't save every state on archive node

magicxyyz:


core/blockchain.go
```diff
@@ -153,8 +156,10 @@ type CacheConfig struct {
 var defaultCacheConfig = &CacheConfig{
 
 	// Arbitrum Config Options
-	TriesInMemory: DefaultTriesInMemory,
-	TrieRetention: 30 * time.Minute,
+	TriesInMemory:                      DefaultTriesInMemory,
+	TrieRetention:                      30 * time.Minute,
+	MaxNumberOfBlocksToSkipStateSaving: 127,
+	MaxAmountOfGasToSkipStateSaving:    15 * 1000 * 1000,
```
tsahee, line 162: Defaults in geth should keep geth (ethereum) behavior as much as possible.
These two should definitely be 0.
Defaults in nitro can be debated (I tend for 0 there too but that's different)

PlasmaPower: On the nitro side, I think trying to match the amount of gas inbetween saved states as L1 (15 million gas) would be good. Saving state for every Arbitrum block isn't as reasonable as it is for L1, but saving it every 15 million gas would mean that the rate of archive disk growth in terms of chain speed should be the same.

PlasmaPower: Then again, this would be something that'd be very easy to miss in an upgrade and would mean that the node is now somewhat permanently missing state. So maybe we should keep it at 0 for the upgrade, I'm not sure. We can talk about it next standup.

tsahee (dismissed):

core/blockchain.go
```diff
@@ -1441,7 +1449,29 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 	}
 	// If we're running an archive node, always flush
```
joshuacolvin0, line 1450: Probably good to update comment to reflect changed code

magicxyyz: I've added a comment, I am open for any suggestions for improvement

joshuacolvin0 (approved):
LGTM

PR #211, opened 2023-04-20
Set IsShanghai true if the ArbOS version is >= 11

amsanghi:


core/vm/runtime/env.go
```diff
@@ -17,11 +17,16 @@
...
 import (
+	"fmt"
+
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/vm"
 )
 
 func NewEnv(cfg *Config) *vm.EVM {
+	if cfg.ChainConfig.IsArbitrum() {
+		panic(fmt.Errorf("chain config has arbitrum enabled"))
```
tsahee, line 28: not important and not mandatory - I'd skip the fmt.error, to reduce merge conflicts in the import section.

params/config.go
```diff
@@ -993,7 +993,7 @@ func (c *ChainConfig) Rules(num *big.Int, isMerge bool, timestamp uint64) Rules
 		IsBerlin:         c.IsBerlin(num),
 		IsLondon:         c.IsLondon(num),
 		IsMerge:          isMerge,
-		IsShanghai:       c.IsShanghai(timestamp),
+		IsShanghai:       currentArbosVersion >= 11,
```
tsahee, line 996: We should still have an IsShanghi func since it's public, it should receive both arbos and timestamp, and should check if it's arbitrum (see IsLondon)

tsahee (commented):
I really like this direction. Very clean.
Tiny comments.

core/vm/runtime/env.go
amsanghi: done.

params/config.go
amsanghi: done.

PlasmaPower:
To improve compatibility with non-Arbitrum chains, I think we should modify the start of `DeserializeHeaderExtraInformation` to look like this:
```go
func DeserializeHeaderExtraInformation(header *Header) HeaderInfo {
	if header.BaseFee == nil || header.BaseFee.Sign() == 0 || len(header.Extra) != 32 || header.Difficulty.Cmp(common.Big1) != 0 {
		// imported blocks have no base fee
		// The genesis block doesn't have an ArbOS encoded extra field
		return HeaderInfo{}
	}
```
We require that the difficulty is 1 (as is true of all nitro blocks), and if the header's Extra field is not 32 bytes long, we just return early which means we don't need an error condition. This should simplify this PR and improve compatibility.

amsanghi:
> To improve compatibility with non-Arbitrum chains, I think we should modify the start of `DeserializeHeaderExtraInformation` to look like this:
> 
> ```go
> func DeserializeHeaderExtraInformation(header *Header) HeaderInfo {
> 	if header.BaseFee == nil || header.BaseFee.Sign() == 0 || len(header.Extra) != 32 || header.Difficulty.Cmp(common.Big1) != 0 {
> 		// imported blocks have no base fee
> 		// The genesis block doesn't have an ArbOS encoded extra field
> 		return HeaderInfo{}
> 	}
> ```
> 
> We require that the difficulty is 1 (as is true of all nitro blocks), and if the header's Extra field is not 32 bytes long, we just return early which means we don't need an error condition. This should simplify this PR and improve compatibility.

done.

PlasmaPower (approved):
LGTM

PR #212, opened 2023-05-01
Make the RPC max response batch size editable

PlasmaPower:
Unfortunately this is a global because too much wiring would be required otherwise.


joshuacolvin0 (approved):
LGTM

PR #214, opened 2023-05-08
RPC-Client improvements

tsahee:
This branch improves rpc client
* Add an interface that ethclient uses instead of the struct
*  Remove some custom code that isn't needed thanks to this interface
* Adds a function to check JWT filepath

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -704,6 +704,14 @@ func (n *Node) WSAuthEndpoint() string {
 	return "ws://" + n.wsAuth.listenAddr() + n.wsAuth.wsConfig.prefix
 }
 
+// JWTPath returns the path for JWT secret
+func (n *Node) JWTPath() string {
+	if n.config.JWTSecret == "" {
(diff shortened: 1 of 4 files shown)


Tristan-Wilson (approved):
LGTM. *Sorry I reviewed this several days ago from the Nitro PR but didn't know there was a separate PR here. My misunderstanding of how github handles submodules.)

node/node.go
```diff
@@ -704,6 +704,14 @@ func (n *Node) WSAuthEndpoint() string {
 	return "ws://" + n.wsAuth.listenAddr() + n.wsAuth.wsConfig.prefix
 }
 
+// JWTPath returns the path for JWT secret
+func (n *Node) JWTPath() string {
+	if n.config.JWTSecret == "" {
+		return n.ResolvePath(datadirJWTKey)
```
PlasmaPower, line 710: IMO, we should use a JWT token outside of the `nitro` directory, so that database copies and snapshots don't also copy the token.

PR #218, opened 2023-05-11
Merge v1.11.3

Tristan-Wilson:
# Notes for initial merge commit and resolution
* setHeadBeyondRoot call site had minor change of header.NumberU64() to head.Number.Uint64(), was missing arbitrum SnapshotRestoreMaxGas and diskRootFound return. Added it back in.
* Many minor changes of header.NumberU64() to head.Number.Uint64().
* ethapi.Backend changed 	CurrentBlock() return from *types.Block to *types.Header, updated NodeInterfaceBackendAPI
* types.TxData interface adds effectiveGasPrice, placeholders currently for arb txs.
* flatCallTracer is missing various arbitrum tracers.Tracer methods, added
stubs.
See comments on subsequent commits for other changes.

# Notes on geth unit tests
```
 Fail on vanilla v1.11.3
 --- FAIL: TestSequentialRead (0.00s)
 --- FAIL: TestSequentialReadByteLimit (0.00s)
 --- FAIL: TestFreezerReadonly (0.00s)
 --- FAIL: TestRandom (0.00s)

 Fails expectedly due to our modifications to Tree.Cap function
 FAIL    github.com/ethereum/go-ethereum/core/rawdb      0.154s
 --- FAIL: TestPostCapBasicDataAccess (0.00s)

 Already was failing on our branch.
 FAIL
 FAIL    github.com/ethereum/go-ethereum/core/state/snapshot     3.550s
 --- FAIL: TestCallTracerNative (0.00s)
     --- FAIL: TestCallTracerNative/simpleOnlytop (0.00s)
 --- FAIL: TestCallTracerNativeWithLog (0.00s)
     --- FAIL: TestCallTracerNativeWithLog/simple (0.00s)
     --- FAIL: TestCallTracerNativeWithLog/calldata (0.00s)
     --- FAIL: TestCallTracerNativeWithLog/withOnlyTopCall (0.00s)
     --- FAIL: TestCallTracerNativeWithLog/txPartialFailed (0.00s)
     --- FAIL: TestCallTracerNativeWithLog/notopic (0.00s)
     --- FAIL: TestCallTracerNativeWithLog/delegatecall (0.00s)
     --- FAIL: TestCallTracerNativeWithLog/multilogs (0.00s)
     --- FAIL: TestCallTracerNativeWithLog/multiContracts (0.01s)
 --- FAIL: TestZeroValueToNotExitCall (0.00s)
 FAIL    github.com/ethereum/go-ethereum/eth/tracers/internal/tracetest  0.086s
```


Tristan-Wilson:
Upstream introduced the flock dependency https://github.com/ethereum/go-ethereum/pull/26633 and currently this is causing the Nitro build to fail because it seems like various platform specific syscall definitions are not implemented for `GOOS=js`.

```
$ GOOS=js GOARCH=wasm go build -o target/tmp/replay.wasm ./cmd/replay/...
# github.com/gofrs/flock
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:28:30: undefined: syscall.LOCK_EX
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:39:30: undefined: syscall.LOCK_SH
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:57:20: undefined: syscall.Flock
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:67:20: undefined: syscall.Flock
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:97:20: undefined: syscall.Flock
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:97:50: undefined: syscall.LOCK_UN
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:119:29: undefined: syscall.LOCK_EX
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:131:29: undefined: syscall.LOCK_SH
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:151:17: undefined: syscall.Flock
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:151:52: undefined: syscall.LOCK_NB
../../go/pkg/mod/github.com/gofrs/flock@v0.8.1/flock_unix.go:151:52: too many errors
```

I haven't decided what to do yet. One of the options would be to roll back the introduction of flock in our fork.

Tristan-Wilson:
Looking through the flock source https://github.com/gofrs/flock, there's no implementation which makes sense for `GOOS=js`, so I'll see about removing the dependency from our fork.

Tristan-Wilson:
This fails with a similar error if I revert 09a9ccdbce88053fb8060d0f4fc82b24a88e3c3a, which uses tsdb

```
GOOS=js GOARCH=wasm go build -o target/tmp/replay.wasm ./cmd/replay/...
package github.com/offchainlabs/nitro/cmd/replay
        imports github.com/ethereum/go-ethereum/consensus
        imports github.com/ethereum/go-ethereum/core/rawdb
        imports github.com/prometheus/tsdb/fileutil
        imports golang.org/x/sys/unix: build constraints exclude all Go files in /home/tristan/go/pkg/mod/golang.org/x/sys@v0.6.0/unix
```

Tristan-Wilson:
Looks like we already fixed this in https://github.com/OffchainLabs/go-ethereum/pull/19 and I accidentally removed it during the merge, will bring it back in.

core/types/arb_types.go
```diff
@@ -100,6 +100,13 @@ func (tx *ArbitrumUnsignedTx) setSignatureValues(chainID, v, r, s *big.Int) {
 
 }
 
+func (tx *ArbitrumUnsignedTx) effectiveGasPrice(dst *big.Int, baseFee *big.Int) *big.Int {
```
tsahee, line 103: For all these specifically I'd like  @PlasmaPower to review

core/rawdb/fileutil.go
```diff
@@ -19,12 +19,13 @@
 
 package rawdb
 
-import "github.com/prometheus/tsdb/fileutil"
+import "github.com/gofrs/flock"
```
tsahee, line 22: this LGTM, but @PlasmaPower should also review

core/types/receipt.go
```diff
@@ -411,25 +411,31 @@ func (rs Receipts) DeriveFields(config *params.ChainConfig, hash common.Hash, nu
...
+
 		// block location fields
 		rs[i].BlockHash = hash
 		rs[i].BlockNumber = new(big.Int).SetUint64(number)
 		rs[i].TransactionIndex = uint(i)
 
 		if rs[i].Type != ArbitrumLegacyTxType {
 			// The contract address can be derived from the transaction itself
-			if txs[i].To() == nil && rs[i].ContractAddress == (common.Address{}) {
+			if txs[i].To() == nil {
```
tsahee, line 423: We don't want to (and possibly cannot) recalculate ContractAddress if it is already filled (by decodeStoredReceiptRLP),

tsahee (commented):
mostly LGTM - I highlighted two areas for review by @PlasmaPower, and one comment in receipt.go

core/types/receipt.go
tsahee: ah, I see this doesn't happen for legacy Tx so I think we always can recalculate it.. question is if we always should

Tristan-Wilson: I added the check back in, thanks for pointing this out.

Tristan-Wilson: I get this error from a nitro test when adding the check back in:
```
$ go test -run "TestContractTxDeploy" ./system_tests
...
    contract_tx_test.go:102:  [expected address 0x0000000000000000000000000000123412341234 nonce 0 to deploy to 0x733A0385EB916548E1a2087F038f0dE75043bEF7 but got 0x0000000000000000000000000000000000000000] 
```

PlasmaPower: This check is already inside this if statement, so adding it back into the upper `if` statement would change behavior by causing it to set `ContractAddress` to empty if it was previously set. So I think this code is good as-is.

PlasmaPower: ah yeah, I see that inner if statement was added recently. Makes sense :)

core/types/arb_types.go
PlasmaPower: These LGTM as they match behavior with the `gasFeeCap()` and `gasTipCap()` impls which say that these txs only specify a fee cap and have a tip cap of zero. That means they'll always just pay the basefee.

core/rawdb/fileutil.go
PlasmaPower: LGTM -- longer term we might want to move the FileLock interface to a common file between this and `filelock_mock.go` but no need to make that change now

PlasmaPower (approved):
LGTM

PR #223, opened 2023-05-24
set evm.Context.Random to be equal to evm.Context.Difficulty

amsanghi:


amsanghi:
> Let's not modify this for non-Arbitrum chains. Let's also remove the changes in cmd, tests, and *_test.go because those won't be used by nitro. I think that leaves two changes, one in core/evm.go and the other in internal/ethapi/api.go. For both of those, let's gate this change behind the ArbOS version being greater than zero (i.e. this is an Arbitrum chain).

done.

PlasmaPower (approved):
LGTM

PR #233, opened 2023-06-12
Fix nonce checks being enabled for Arbitrum "fake" txs

PlasmaPower:
Transactions such as L1->L2 contract transactions should bypass nonce checks, as they previously did via IsFake, which has been renamed to SkipAccountChecks in recent geth versions.


tsahee (approved):
LGTM

PR #234, opened 2023-06-14
cherry pick gas estimation using latest block

anodar:
Cherry pick: https://github.com/ethereum/go-ethereum/commit/0b66d47449f61e9ebaf9e1db3ed290b59844d4c1 commit, that uses latestBlockNumber for gas estimation instead of pending.


PlasmaPower (approved):
LGTM

PR #236, opened 2023-06-20
Pin the block for binary search in gas estimation

anodar:
Also:

- use option structure [0] to make the code more readable
- factor out feeCap in separate helper for simplicity
- inline error handling (if err := ..; err != nil)

[0] https://google.github.io/styleguide/go/best-practices#option-structure


anodar:
Closing this as upstream already fixed this issue in https://github.com/ethereum/go-ethereum/commit/8c288b528d3c9653c84321387e78687ed02134d7

PR #237, opened 2023-06-22
Fix EVM for gas estimation for scheduled messages

PlasmaPower:
The EVM was previously being created for the original message, not the scheduled message which is actually being executed with the EVM


Tristan-Wilson (approved):
LGTM

PR #238, opened 2023-06-28
Execution split support

tsahee:


arbitrum/apibackend.go
```diff
@@ -86,21 +86,28 @@ type SyncProgressBackend interface {
...
 		fallbackClient: fallbackClient,
-		sync:           sync,
 	}
 	filterSystem := filters.NewFilterSystem(backend.apiBackend, filterConfig)
 	backend.stack.RegisterAPIs(backend.apiBackend.GetAPIs(filterSystem))
 	return filterSystem, nil
 }
 
+func (a *APIBackend) SetSyncBackend(sync SyncProgressBackend) error {
+	if a.sync != nil && sync != nil {
```
PlasmaPower, line 104: Is this condition designed to allow unsetting the sync backend? Why would we want to do that?

PlasmaPower (commented):
LGTM (though we shouldn't merge this until we're ready to merge the nitro side, because this is an API breaking change)

PlasmaPower (approved):
LGTM

PR #240, opened 2023-07-07
recordingdb: add config, trie flush, and metrics

tsahee:


PlasmaPower (commented):
LGTM, but let's not merge this until everything else is ready because this is a breaking API change for nitro

PlasmaPower (approved):
LGTM

PR #241, opened 2023-07-12
Default path of nodekey is updated to be outside the nitro directory

ganeshvanahalli:
`$GlobalConfig = ~/.arbitrum` by default
`$Chain = $GlobalConfig/arb1` for the arb1 chain by default

For nodekey, `current path = $Chain/nitro/nodekey` and `updated path = $Chain/nodekey`


node/config.go
```diff
@@ -394,7 +400,7 @@ func (c *Config) NodeKey() *ecdsa.PrivateKey {
 	if err != nil {
 		log.Crit(fmt.Sprintf("Failed to generate node key: %v", err))
 	}
-	instanceDir := filepath.Join(c.DataDir, c.name())
+	instanceDir := c.DataDir
```
Tristan-Wilson, line 403: It looks like it used to be that `keyfile` for LoadECDSA and for `SaveECDSA` were different due to the inclusion of `c.name()` in the `instanceDir`, and now it can be the same. I don't know the history behind this or what the behavior should be.

node/config.go
```diff
@@ -385,7 +385,13 @@ func (c *Config) NodeKey() *ecdsa.PrivateKey {
 		return key
 	}
 
-	keyfile := c.ResolvePath(datadirPrivateKey)
+	// Updating default path to nodekey to be $Chain/nodekey
+	var keyfile string
+	if filepath.IsAbs(datadirPrivateKey) {
```
Tristan-Wilson, line 390: Do we need to handle absolute paths for when we're setting `keyfile` for the `SaveECDSA` code below?

node/config.go
ganeshvanahalli: Actually, keyfile's directory and instanceDir were same before too. 
This is ensured by keyfile being assigned to`ResolvePath`'s return value i.e  `filepath.Join(c.instanceDir(), path)`

Tristan-Wilson (approved):
LGTM

PR #242, opened 2023-07-17
add AuthModules and AuthOrigins params to stack config

magicxyyz:
* adds AuthModules and AuthOrigins to node.Config
* previously, only option to set those was to change global defaults


Tristan-Wilson (approved):
LGTM

PR #243, opened 2023-07-25
Export errNotFound from memorydb to use it in Nitro error checks

anodar:


PlasmaPower:
To minimize our geth diff and future work merging in upstream geth releases, could we only export ErrMemorydbNotFound and not the other ones?

anodar:
Sounds good, reverted other errors.

PlasmaPower (approved):
LGTM

PR #251, opened 2023-08-28
Immediately return an error from updateTrie if DeleteStorage fails

PlasmaPower:
The code modified in this PR is the deterministic trie deletion code which we added. Upstream go-ethereum modified the non-deterministic case to immediately return an error if DeleteStorage fails, so we should do the same in the deterministic case.


Tristan-Wilson (approved):
LGTM

PR #252, opened 2023-08-30
Upgrade ConditionalOptions variables to type math.HexOrDecimal64 to support eth_sendRawTransactionConditional

ganeshvanahalli:


common/math/integer_test.go
```diff
@@ -114,3 +115,78 @@ func TestMustParseUint64Panic(t *testing.T) {
...
+		{input: `"10"`, want: uint64(10)},
+		{input: `"100"`, want: uint64(100)},
+		{input: `"12344678"`, want: uint64(12344678)},
+		{input: `"1111111111111111"`, want: uint64(1111111111111111)},
+		{input: `"0x0"`, want: uint64(0)},
+		{input: `"0x2"`, want: uint64(0x2)},
+		{input: `"0x2F2"`, want: uint64(0x2f2)},
+		{input: `"0x1122aaff"`, want: uint64(0x1122aaff)},
+		{input: `"0xbbb"`, want: uint64(0xbbb)},
+		{input: `"0xffffffffffffffff"`, want: uint64(0xffffffffffffffff)},
```
PlasmaPower, line 157: I think the hex based inputs should be `"\"0x2\""` for example, as the strings are supposed to be JSON, but this doesn't matter with the current implementation of HexOrDecimal64. A follow-up PR to fix this might still be good :)

PlasmaPower (approved):
LGTM -- I'm going to merge this despite my comment on the test so we can get this into the next beta release

common/math/integer_test.go
PlasmaPower: That, or we could call UnmarshalText instead of UnmarshalJSON in the test

ganeshvanahalli: Agreed, the previous tests covered UnmarshalText so I wanted to test for JSON too, I'll put a followup PR to fix the inputs

PR #259, opened 2023-09-28
Fix error handling of batch pushResponse

PlasmaPower:
https://github.com/OffchainLabs/go-ethereum/pull/258 will eventually have a better fix, but in the meantime, this fixes the error handling by using the higher level `callBuffer` instead of `h.conn.writeJSON`


magicxyyz (approved):
LGTM

PR #260, opened 2023-10-02
Add tipping tx subtype

magicxyyz:
* adds new transaction type which allows creating new subtypes of the transaction (`ArbitrumSubtypedTx`)
* adds new transaction subtype that wraps DynamicFeeTx and supports collection of the priority fee (`ArbitrumTippingTx`)


core/types/transaction.go
```diff
@@ -128,6 +128,11 @@ func (tx *Transaction) EncodeRLP(w io.Writer) error {
 
 // encodeTyped writes the canonical encoding of a typed transaction to w.
 func (tx *Transaction) encodeTyped(w *bytes.Buffer) error {
+	if tx.Type() == ArbitrumSubtypedTxType {
+		w.WriteByte(tx.Type())
+		w.WriteByte(tx.inner.(*ArbitrumSubtypedTx).TxSubtype())
+		return rlp.Encode(w, tx.inner.(*ArbitrumSubtypedTx).TxData)
+	}
```
magicxyyz, line 135: we need this change here to be able to decode arbitrary internal types that implement TxData (similarly as it is done for top level `Transaction`)

internal/ethapi/api.go
```diff
@@ -1625,6 +1626,25 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
...
+	case *types.ArbitrumSubtypedTx:
+		subtype := inner.TxSubtype()
+		result.Subtype = hexutil.Uint64(subtype)
+		switch subtype {
+		case types.ArbitrumTippingTxSubtype:
+			al := tx.AccessList()
+			result.Accesses = &al
+			result.ChainID = (*hexutil.Big)(tx.ChainId())
+			result.GasFeeCap = (*hexutil.Big)(tx.GasFeeCap())
+			result.GasTipCap = (*hexutil.Big)(tx.GasTipCap())
```
tsahee, line 1638: don't you also need to set result.YParity?

core/types/arbitrum_signer.go
```diff
@@ -40,6 +41,20 @@ func (s arbitrumSigner) Sender(tx *Transaction) (common.Address, error) {
 		}
 		fakeTx := NewTx(&legacyData.LegacyTx)
 		return s.Signer.Sender(fakeTx)
+	case *ArbitrumSubtypedTx:
+		switch inner.TxData.(type) {
+		case *ArbitrumTippingTx:
+			V, R, S := tx.RawSignatureValues()
```
tsahee, line 47: 3 options, I'm o.k. with any:
1) add a comment that this comes from london signer as these are assentially DynamicFeeTxType
2) create a fake DynamicFeeTx and use the internal signer, similar to ArbitrumLegacyTx
3) separate the code that extract the sender inside london-signer (without checking tx type), call this func from both london signer after checking type and from here.

tsahee (commented):
a few cpmments.
overall - looking good

core/types/transaction.go
```diff
@@ -244,6 +244,10 @@ func (tx *Transaction) decodeTyped(b []byte, arbParsing bool) (TxData, error) {
 		var inner BlobTx
 		err := rlp.DecodeBytes(b[1:], &inner)
 		return &inner, err
+	case ArbitrumSubtypedTxType:
```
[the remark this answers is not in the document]
magicxyyz: we need to parse `ArbitrumSubtypedTxType` even when `arbParsing` is false to be able to decode binary encoding with `UnmarshalBinary`  to support `SendRawTransaction` and `SendRawTransactionConditional`. If we don't need to support it, then we probably need to add custom RPC method to allow sending subtyped txes (?).

core/types/transaction_marshalling.go
```diff
@@ -48,6 +48,7 @@ type txJSON struct {
 	YParity              *hexutil.Uint64 `json:"yParity,omitempty"`
 
 	// Arbitrum fields:
+	Subtype             *hexutil.Uint64 `json:"subtype,omitempty"`             // ArbitrumSubtypedTx
```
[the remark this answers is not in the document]
magicxyyz: 1. to make it consistent with type of "type" field which is also `hexutil.Uint64`
2. there's no `hexutil.Byte`, and `hexautil.Uint64` does all format checking and decoding from string

internal/ethapi/api.go
magicxyyz: good catch! thanks

core/types/arbitrum_signer.go
magicxyyz: 2 - we can't make fake `DynamicFeeTx` and pass it to `londonSigner` because `Hash` method must be called on `arbitrumSigner` to include transaction subtype in hashed data
3 - probably would create some unnecessary merge conflicts against upstream

so I went with 1 and added a comment. 

In `SignatureValues` method I used the fake tx pattern, as there it was possible.

core/types/arb_types.go
```diff
@@ -516,3 +525,91 @@ func DeserializeHeaderExtraInformation(header *Header) HeaderInfo {
...
+}
+
+func (tx *ArbitrumSubtypedTx) copy() TxData {
+	return &ArbitrumSubtypedTx{
+		TxData: tx.TxData.copy(),
+	}
+}
+
+func (tx *ArbitrumSubtypedTx) txType() byte    { return ArbitrumSubtypedTxType }
+func (tx *ArbitrumSubtypedTx) TxSubtype() byte { return tx.TxData.txType() }
```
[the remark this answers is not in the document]
magicxyyz: I'd be great to keep the subtype a constant and adduct a field invalidates the property

Maybe I should add a `func (tx *ArbitrumSubtypedTx) SubTx() TxData` method, and then we'd use it this way: `tx.SubTx().Type()`, would it be better?

magicxyyz:
we went with a different approach for now

PR #261, opened 2023-10-05
garbage collect not commited tries

magicxyyz:
* adds missing garbage collection of not committed states when running archive node that skips saving some states
* adds acquiring db lock  before accessing dirties map in `hashdb/Database.commit` to fix race `Database.commit` vs `Database.dereference`
* fixes committing oldest node from triegc in BlockChain.Stop


ganeshvanahalli (approved):
lgtm

PR #269, opened 2023-11-04
Implement more complex slot value comparisons

diwu1989:
Add an optional `SlotValueComparison` to support `=,!=,>,<` value comparisons in conditional TX.
The current equality check is not sufficient for detecting specific kinds of value changes in a slot value.

### DIFF
--- a/arbitrum_types/txoptions.go
+++ b/arbitrum_types/txoptions.go
@@ -46,9 +46,17 @@ func WrapOptionsCheckError(err error, msg string) error {
 	}
 }
 
+type SlotValueComparison struct {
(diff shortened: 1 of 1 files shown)


hkalodner:
@cla-bot check

joshuacolvin0:
The only reason `eth_sendRawTransactionConditional` exists is to provide compatibility with account abstraction bundlers.  The implementation follows the API designed for any L2, and is described here: https://notes.ethereum.org/@yoav/SkaX2lS9j

The functionality is limited on purpose to reduce DOS risk.

PR #270, opened 2023-11-16
fix closing TrieDB in eth.StateAtBlock

magicxyyz:
This PR fixes closing triedb in `eth.StateAtBlock`. Previously `trie.Database.Dereference` was called in `tracers.StateReleaseFunc` after the db was closed in deferred function, what wasn't a correct behavior.


joshuacolvin0 (approved):
LGTM

PR #271, opened 2023-11-21
Revert "core/state: mark account as dirty when resetObject occurs"

PlasmaPower:
This reverts commit 15bd21f3c878155bc2254bb43460763298f58ad1.

This worked for upstream go-ethereum, but some ways we manipulate the state means this changes behavior for us

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -180,6 +180,12 @@ func (ch resetObjectChange) revert(s *StateDB) {
 }
 
 func (ch resetObjectChange) dirtied() *common.Address {
+	// Arbitrum: We keep the behavior that existed before go-ethereum v1.12.1 and return nil,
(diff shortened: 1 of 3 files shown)


joshuacolvin0 (approved):
LGTM

tsahee:
seems fine.. but need to understand that test failure (I re-ran it to see if it's just flakiness)

PlasmaPower:
Hmm yeah this seems to have broken something. I'll take a look tomorrow

PlasmaPower:
The failing test is in `state_fuzz_test.go` and they say "We're mostly adding this change to appease the fuzzer" so I think it might be intended and we should just disable that test.

PlasmaPower:
I've skipped that test. I've been syncing a node overnight with this change and it agrees with the rest of the network, so I think it's working (definitely working better than the previous alpha).

PR #273, opened 2023-11-21
ethdb/pebble: don't double-close iterator inside pebbleIterator

magicxyyz:
Adds released flag to pebbleIterator to avoid double closing cockroachdb/pebble.Iterator as it is an invalid operation.

This PR backports: https://github.com/ethereum/go-ethereum/pull/28566
Fixes: https://github.com/ethereum/go-ethereum/issues/28565


PlasmaPower (approved):
LGTM

PR #276, opened 2023-12-06
fix prunning: use correct state root and don't require account address

magicxyyz:
This PR:
* uses block root as state root in `trie.StorageTrieID` call (previously leaf key was passed)
 https://github.com/OffchainLabs/go-ethereum/blob/b1622e6ac4bf3762aebde92a585de2889d90823f/trie/trie_id.go#L22-L26
 https://github.com/OffchainLabs/go-ethereum/blob/b1622e6ac4bf3762aebde92a585de2889d90823f/trie/trie_id.go#L37-L45
 
 * similarly to upstream, uses `trie.NewStateTrie` directly instead of `OpenStorageTrie` to not require preimage of account address hash


PlasmaPower (approved):
LGTM

PR #277, opened 2023-12-07
use StateAtBlock and reference states when recreating missing state

magicxyyz:
This PR is addressing following issues:

1. When getting state for RPCs (in the upstream code), the state getter is called first and after that the state root is referenced in `hashdb` dirties cache if at all. It means that the state could potentially be garbage collected in another thread and further operations on state would fail.
That is fine for normal archive node that keeps all the state and doesn't use dirties cache. It is also ok for a full node as we don't require it to keep all the state but only some recent, so RPCs are expected to fail when requesting too old state. 
However, that is an issue for "hybrid" node that persists only some states and keeps recent states in dirties cache - we expect it to always be able to process RPC for any state (to emulate normal archive node functionality).
2. Previous implementation of recreating missing state (`AdvanceStateUpToBlock`), which we use for some RPCs (e.g. `eth_call`, `etc_getBalance`), accumulated changes to state in one `StateDB` object, what may cause some issues i.a. excessive memory usage by caches inside `StateDB`.

This PR:
1. reorders referencing and getting state, adds referencing  and adds a release method to `state.StateDB`
2. changes the recreation of state to use `eth.StateAtBlock` which is upstream implementation for recreation of state for tracers and should be a more robust solution


arbitrum/apibackend.go
```diff
@@ -433,40 +434,59 @@ func (a *APIBackend) BlockByNumberOrHash(ctx context.Context, blockNrOrHash rpc.
 	return nil, errors.New("invalid arguments; neither block nor hash specified")
 }
 
-func (a *APIBackend) stateAndHeaderFromHeader(ctx context.Context, header *types.Header, err error) (*state.StateDB, *types.Header, error) {
+func (a *APIBackend) stateAndHeaderFromHeader(ctx context.Context, header *types.Header, err error) (*state.StateDB, *types.Header, ethapi.StateReleaseFunc, error) {
```
tsahee, line 437: I see why this solution makes sense and it's probably what I'd go for if we were maintaining geth.. but this seems like a merge hell. Maybe it's required but I'm thinking about ways to avoid it..

should we use a finalizer for the returned statedb instead? (not sure this is a good idea, as it's ugly and trusting geth itself will never set a finalizer for it)

tsahee (commented):
Generally looking good. Some comments.
Also - I'd like to have a test for our partial-archive-node config that will try at least some of these APIs.

arbitrum/apibackend.go
magicxyyz: let's discuss it on the meeting tommorrow, we could go with the finalizer, but what worries me is that there is no guarantee when it will be called and that could be an issue during congestion of RPC calls:

```
// Note that because finalizers may execute arbitrarily far into the future
// after an object is no longer referenced, the runtime is allowed to perform
// a space-saving optimization that batches objects together in a single
// allocation slot. The finalizer for an unreferenced object in such an
// allocation may never run if it always exists in the same batch as a
// referenced object. Typically, this batching only happens for tiny
// (on the order of 16 bytes or less) and pointer-free objects.
```
(https://github.com/golang/go/blob/9db508d46fe201ecc6184ebd7b73770ed38fbdf9/src/runtime/mfinal.go#L369-L376)

Seems that it should "typically" work, as `StateDB` has pointer fields, imo we have to decide whether it is good enough.

magicxyyz:
> Also - I'd like to have a test for our partial-archive-node config that will try at least some of these APIs.

There already is a system test on nitro side, that runs this partial-archive-node, stops it, restarts it and then checks if `eth_getBalance` can be called for some blocks. 
https://github.com/OffchainLabs/nitro/blob/a40f8f1e9ba0975452d3dc1da602f4e20142b608/system_tests/recreatestate_rpc_test.go#L429-L429

If needed, I can add some more specific tests for the changes, also in geth repo.

magicxyyz:
> > Also - I'd like to have a test for our partial-archive-node config that will try at least some of these APIs.
> 
> There already is a system test on nitro side, that runs this partial-archive-node, stops it, restarts it and then checks if `eth_getBalance` can be called for some blocks. https://github.com/OffchainLabs/nitro/blob/a40f8f1e9ba0975452d3dc1da602f4e20142b608/system_tests/recreatestate_rpc_test.go#L429-L429
> 
> If needed, I can add some more specific tests for the changes, also in geth rep

As suggested, I added a specific test on nitro side for getting state for RPCs:
https://github.com/OffchainLabs/nitro/blob/bb5c908d7c16e103130e2d80c7f5fc01e4dbef2e/system_tests/recreatestate_rpc_test.go#L514C1-L557

arbitrum/apibackend.go
```diff
@@ -445,20 +446,57 @@ func (a *APIBackend) stateAndHeaderFromHeader(ctx context.Context, header *types
 	}
 	bc := a.BlockChain()
 	stateFor := func(header *types.Header) (*state.StateDB, error) {
```
tsahee, line 448: because StateFor is only used internaly and in two places, I think it's safe to have it return a release function and use it.
FindLastAvailableState can also return the release function for latest available state.
Then we'll only set a finalizer in stateAndHeaderFromHeader

core/state/statedb.go
```diff
@@ -145,9 +145,28 @@ type StateDB struct {
...
+	released bool
+
 	deterministic bool
 }
 
+func (s *StateDB) SetRelease(release func()) {
+	if s.release == nil {
+		s.release = release
+	} else {
+		log.Error("StateDB.SetRelease called more then once, may cause memory leak")
```
tsahee, line 159: Probably better to return an error from SetRelease in that case (and check it in callers)
I want to have something that would make tests fail.

core/state/statedb.go
```diff
@@ -145,9 +145,28 @@ type StateDB struct {
...
 	onCommit func(states *triestate.Set) // Hook invoked when commit is performed
 
+	// arbitrum: cleanup hook
+	release  func()
+	released bool
+
 	deterministic bool
 }
 
+func (s *StateDB) SetRelease(release func()) {
```
tsahee, line 155: Let's talk about why finalizers don't work.
If not finalizers - I think I prefer your original solution returning finalizers when creating StateDB.
When upstream changes - it's best if we don't have to do anything (which was the idea of finalizers). Returning a new param will make us have to fix many merge conflicts which is a pain.
However, current approach could be worse - if some new place calls StateAtBlock we'll probably just not notice it, not add a defer state.Release where needed and end up with a memory leak.

eth/state_accessor.go
```diff
@@ -70,7 +80,7 @@ func (eth *Ethereum) hashState(ctx context.Context, block *types.Block, reexec u
 			database = state.NewDatabaseWithConfig(eth.chainDb, trie.HashDefaults)
 			if statedb, err = state.New(block.Root(), database, nil); err == nil {
 				log.Info("Found disk backend for state trie", "root", block.Root(), "number", block.Number())
-				return statedb, func() { database.TrieDB().Close() }, nil
+				return statedb, noopReleaser, nil
```
tsahee, line 83: why is that changed to noop? where will that temporarye stateDatabase be released?

eth/state_accessor.go
```diff
@@ -93,7 +103,7 @@ func (eth *Ethereum) hashState(ctx context.Context, block *types.Block, reexec u
 		if !readOnly {
 			statedb, err = state.New(current.Root(), database, nil)
 			if err == nil {
-				return statedb, func() { database.TrieDB().Close() }, nil
+				return statedb, noopReleaser, nil
```
tsahee, line 106: same question

eth/state_accessor.go
magicxyyz: 1. Referencing here is not needed, as if the state is available it was read from disk.

2. I've removed some of our changes here to minimize the diff to upstream, ia. we went with how upstream solved cleans cache memory leak by just disabling the cleans cache for the temporary dbs (trie.HashDefaults has cleans = 0) => we don't need to close the db here. If there will be need for cleans cache for better performance, we can enable cleans cache - we can do that in separate PR, but if needed I can enable it in this PR.

eth/state_accessor.go
magicxyyz: here also if state is available then it was read from disk + same as above we don't need to close the triedb so we can remove the diff.

tsahee (dismissed):
LGTM

tsahee (commented):
LGTM

tsahee (approved):
I meant approve, LGTM:)

PR #278, opened 2024-01-16
Increase classic L1 block search limit

PlasmaPower:
There only seems to be a maximum gap of 8 classic blocks without transactions, but this PR increases the limit to 50 to be safe.


joshuacolvin0:
LGTM

joshuacolvin0 (approved):
LGTM

PR #281, opened 2024-01-22
Add components that geth backend has in arbitrum's backend

ganeshvanahalli:


arbitrum/backend.go
```diff
@@ -32,6 +35,8 @@ type Backend struct {
 	chanTxs      chan *types.Transaction
 	chanClose    chan struct{} //close coroutine
 	chanNewBlock chan struct{} //create new L2 block unless empty
+
```
magicxyyz, line 38: just a note: I wonder if we should add also `eth.handler` component here.
https://github.com/OffchainLabs/go-ethereum/blob/d3bcf56486451690fcb366a6cefa3a74429e362c/eth/backend.go#L75

 It is connected to the p2p protocols => I believe we will need that when adding snap-sync support, but we might as well just add this field then.

magicxyyz (approved):
LGTM

PR #286, opened 2024-01-31
Update blockContext's time field for pending blocks to match current time. Enable eth_estimateGas and eth_call rpcs to utilize this

ganeshvanahalli:


internal/ethapi/api.go
```diff
@@ -1349,7 +1357,7 @@ func DoEstimateGas(ctx context.Context, b Backend, args TransactionArgs, blockNr
 
 	// We first execute the transaction at the highest allowable gas limit, since if this fails we
 	// can return error immediately.
-	failed, result, err := executeEstimate(ctx, b, args, state.Copy(), header, vanillaGasCap, hi)
+	failed, result, err := executeEstimate(ctx, b, args, state.Copy(), header, vanillaGasCap, hi, &blockNrOrHash)
```
PlasmaPower, line 1360: Why not update header.Time here instead? Btw, we don't want different executions of the estimate to run against different timestamps, which would be an issue with this implementation.

PlasmaPower (changes requested):

internal/ethapi/api.go
ganeshvanahalli: Nice catch! I prioritized updating block context's time over header's but like you said it shouldn't matter for pending blocks. I'll fix this.

internal/ethapi/api.go
```diff
@@ -1165,13 +1165,28 @@ func doCall(ctx context.Context, b Backend, args TransactionArgs, state *state.S
...
 
+func updateHeaderForPendingBlocks(blockNrOrHash rpc.BlockNumberOrHash, header *types.Header) *types.Header {
+	if blockNrOrHash.BlockNumber != nil &&
+		*blockNrOrHash.BlockNumber == rpc.PendingBlockNumber {
+		headerCopy := *header
+		blkTime := uint64(time.Now().Unix())
+		if blkTime > headerCopy.Time {
+			headerCopy.Time = blkTime
+		}
+		headerCopy.Number = big.NewInt(headerCopy.Number.Int64() + 1)
```
PlasmaPower, line 1176: Let's do `headerCopy.Number = new(big.Int).Add(headerCopy.Number, common.Big1)`, which is a bit more efficient

internal/ethapi/api.go
```diff
@@ -1222,8 +1237,8 @@ func (e *revertError) ErrorData() interface{} {
 // useful to execute and retrieve values.
 func (s *BlockChainAPI) Call(ctx context.Context, args TransactionArgs, blockNrOrHash *rpc.BlockNumberOrHash, overrides *StateOverride, blockOverrides *BlockOverrides) (hexutil.Bytes, error) {
 	if blockNrOrHash == nil {
-		latest := rpc.BlockNumberOrHashWithNumber(rpc.LatestBlockNumber)
-		blockNrOrHash = &latest
+		pending := rpc.BlockNumberOrHashWithNumber(rpc.PendingBlockNumber)
+		blockNrOrHash = &pending
```
PlasmaPower, line 1241: For eth_call I think we still want to default to `latest` to maintain improved compatibility with L1 (though if a user specifies "pending" we also want to support that)

PlasmaPower (approved):
LGTM

PR #287, opened 2024-02-01
Parse ABI errors in abi/bind

PlasmaPower:
Newer Solidity versions support returning ABI encoded errors, instead of the old string errors. This PR updates abi/bind to parse them, and provide nice errors like `execution reverted: NoSuchKeyset(0x0100000000000000000000000000000000000000000000000000000000000000)`

This can't decode the errors of anything called by the target contract unless it also includes those errors, but this should already be a big improvement.

This PR also updates the incorrect (or rather, too vague) return type of the Error Unpack method.


joshuacolvin0 (approved):
LGTM

PR #288, opened 2024-02-06
Fix ArbitrumLegacyTxData decoding

PlasmaPower:
ArbitrumLegacyTxData is RLP encoded and decoded, but the fallback to LegacyTxData's implementation of encode and decode was causing panics: https://github.com/OffchainLabs/nitro/issues/2131

### DIFF
--- a/core/types/arbitrum_legacy_tx.go
+++ b/core/types/arbitrum_legacy_tx.go
@@ -49,6 +49,14 @@ func (tx *ArbitrumLegacyTxData) copy() TxData {
 
 func (tx *ArbitrumLegacyTxData) txType() byte { return ArbitrumLegacyTxType }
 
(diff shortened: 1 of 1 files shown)


joshuacolvin0 (approved):
LGTM

Tristan-Wilson (approved):
LGTM

I checked how we had missed this. In this commit https://github.com/OffchainLabs/go-ethereum/commit/8303a8bdf908403c9e92a6da77a3cafee813e1cf upstream switched over from rlp.DecodeBytes on all the tx types to using their .decode methods, and we didn't have one for ArbitrumLegacyTxType https://github.com/OffchainLabs/go-ethereum/commit/8303a8bdf908403c9e92a6da77a3cafee813e1cf#diff-8fb609018eccff10931645347515b8ecc194992cc4ef9e84ebdea87682dd50c3R218)

PR #290, opened 2024-02-16
Fix UsedGas of scheduling txs for gas estimation

PlasmaPower:
This is important because gas estimation now does an initial run of the transaction to determine a lower gas bound. This gas bound was previously incorrect for transactions that scheduled a retryable redemption.


joshuacolvin0 (approved):
LGTM

PR #291, opened 2024-02-20
backport upstream hashdb locking changes

magicxyyz:
backports: https://github.com/ethereum/go-ethereum/pull/28542

### DIFF
--- a/trie/triedb/hashdb/database.go
+++ b/trie/triedb/hashdb/database.go
@@ -180,7 +180,7 @@ func (db *Database) insert(hash common.Hash, node []byte) {
 
 // node retrieves an encoded cached trie node from memory. If it cannot be found
 // cached, the method queries the persistent database for the content.
-func (db *Database) Node(hash common.Hash) ([]byte, error) {
+func (db *Database) node(hash common.Hash) ([]byte, error) {
 	// It doesn't make sense to retrieve the metaroot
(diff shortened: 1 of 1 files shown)


trie/triedb/hashdb/database.go
```diff
@@ -183,7 +178,7 @@ func (db *Database) insert(hash common.Hash, node []byte) {
 	db.dirtiesSize += common.StorageSize(common.HashLength + len(node))
 }
 
-// Node retrieves an encoded cached trie node from memory. If it cannot be found
+// node retrieves an encoded cached trie node from memory. If it cannot be found
 // cached, the method queries the persistent database for the content.
 func (db *Database) Node(hash common.Hash) ([]byte, error) {
```
tsahee, line 183: In this one you did not make the function private. Is that on purpose? Why?

tsahee: I'm really curious if / how well we can implement RecordingKV based off pathdb. I don't know the details.
I assume it's not trivial, so I'd say for this PR let's have a Node function exporting node, and leave pathdb support for a separate PR.

magicxyyz: makes sense, I've added the Node method, thanks!

tsahee (approved):
LGTM

PR #298, opened 2024-03-28
Fix balance burn error message

rachel-bousfield:
This PR fixes a balance burn error message


joshuacolvin0 (approved):
LGTM

rachel-bousfield:
Closing since `opSelfdestruct6780` now records a burn equal to the amount in the account due to the new `SubBalance` line. The previous behavior should be correct.

PR #301, opened 2024-04-03
core/vm, params: ensure order of forks, prevent overflow (#29023)

Tristan-Wilson:
Closes NIT-2397

This PR fixes an overflow which can could happen if inconsistent blockchain rules were configured. Additionally, it tries to prevent such inconsistencies from occurring by making sure that merge cannot be enabled unless previous fork(s) are also enabled.

(cherry picked from commit ac0ff044606a663eeb47ef60ed5506f842753084) aka https://github.com/ethereum/go-ethereum/pull/29023


joshuacolvin0:
Does not apply to Nitro because exploit requires pre EIP-150 block to trigger issue.  Since EIP-150 has been active since genesis, there are no pre EIP_150 blocks.

This change will be merged with regular upstream merges

PR #303, opened 2024-04-15
Add Precompile for secp256r1 conditionally based on ArbOS version

anodar:
Main changes forked from: https://github.com/ulerdogan/go-ethereum/pull/1/files


core/vm/contracts.go
```diff
@@ -145,13 +153,20 @@ func init() {
...
 }
 
 // ActivePrecompiles returns the precompiles enabled with the current configuration.
 func ActivePrecompiles(rules params.Rules) []common.Address {
 	switch {
 	case rules.IsArbitrum:
-		return PrecompiledAddressesArbitrum
+		var ret []common.Address
+		if rules.ArbOSVersion >= 30 {
+			ret = append(ret, PrecompiledAddressesP256Verify...)
```
PlasmaPower, line 167: - This is called every tx, so I think we should consider caching this as PrecompiledAddressesArbOS30 or such
- We also need to update the EVM's `precompile` method in core/vm/evm.go

anodar: Good catch, updated evm's precompile as well and introduced PrecompiledAddresses[/Contracts]ArbOS30 that is populated in geth hook (https://github.com/OffchainLabs/nitro/pull/2239).

core/vm/contracts.go
```diff
@@ -122,11 +129,12 @@ var PrecompiledContractsBLS = map[common.Address]PrecompiledContract{
 }
 
 var (
-	PrecompiledAddressesCancun    []common.Address
-	PrecompiledAddressesBerlin    []common.Address
-	PrecompiledAddressesIstanbul  []common.Address
-	PrecompiledAddressesByzantium []common.Address
-	PrecompiledAddressesHomestead []common.Address
+	PrecompiledAddressesP256Verify []common.Address
```
magicxyyz, line 132: do we need this? looks like we include those addresses in `PrecompiledAddressesArbOS30` instead

magicxyyz (changes requested):
looks good, just might need to remove `PrecompiledAddressesP256Verify` if it isn't used

core/vm/contracts.go
PlasmaPower: Removed

PR #307, opened 2024-04-22
[NIT-2437] Propogate error as reason while stopping flatTracer

amsanghi:
Thanks @tsahee for figuring out the issue!

### DIFF
--- a/eth/tracers/native/call_flat.go
+++ b/eth/tracers/native/call_flat.go
@@ -248,6 +248,7 @@ func (t *flatCallTracer) GetResult() (json.RawMessage, error) {
 
 // Stop terminates execution of the tracer at the first opportune moment.
 func (t *flatCallTracer) Stop(err error) {
+	t.reason = err
 	t.tracer.Stop(err)
 }
 


eth/tracers/native/call_flat.go
```diff
@@ -248,6 +248,7 @@ func (t *flatCallTracer) GetResult() (json.RawMessage, error) {
 
 // Stop terminates execution of the tracer at the first opportune moment.
 func (t *flatCallTracer) Stop(err error) {
+	t.reason = err
```
tsahee, line 251: I'd go the other way..
remove t.reason here, and return t.tracer.reason from GetResults

amsanghi: Infura has submitted a patch with `t.reason = err ` https://github.com/ethereum/go-ethereum/pull/29623/files should i still update as above or let it be as it is (cause it will cause conflict in future while merging geth, I don't have a preference either way)?

PlasmaPower:
Sorry I didn't see this PR. I made one copy-pasting Infura's new approach (which matches @tsahee's suggestion) here: https://github.com/OffchainLabs/go-ethereum/pull/310

I'll close this one accordingly. Thanks for helping out!

PR #310, opened 2024-04-24
Cherry-pick flat call tracer stopping bug

PlasmaPower:
This is a manual cherry pick of https://github.com/ethereum/go-ethereum/pull/29623 since there was quite a lot of tracer-related diff from https://github.com/ethereum/go-ethereum/pull/29189


joshuacolvin0 (approved):
LGTM

PR #311, opened 2024-04-26
Merge v1.13.11

ganeshvanahalli:
Merge upstream geth's v1.13.11 release into our geth fork:https://github.com/ethereum/go-ethereum/releases/tag/v1.13.11
* Upstream disabled logging by default, but we keep it on since the logs are very useful for tests in nitro
* This [PR](https://github.com/ethereum/go-ethereum/pull/28837) in upstream geth introduces a deadlock when using simulated_beacon_api with multiple transactions being submitted simultaneously (an often situation in multiple nitro tests). We also include a fix for this
  * The api in simulator mode, eventually calls forced reset on txpool, and that calls each of the subpools to notify its subscribers (which is blocking) with transactions that were affected by the reset, and the api is one of the subscribers- thus a deadlock :) 
  * The fix is to make the txEvents channel that api uses a buffered one, here we choose the capacity to be 15 (which is ample for situations handled in nitro tests)
* On M2 mac, metrics/influxdb/ tests fail due to different floating point results

_Uses Tristian's https://github.com/Tristan-Wilson/MyScripts/blob/main/bin/merge-helper_

nitro PR- https://github.com/OffchainLabs/nitro/pull/2259
Part of NIT-2452

# Testing done
geth and nitro tests pass


Tristan-Wilson (approved):
LGTM

PR #312, opened 2024-05-01
Fix requested block ahead of current block check

PlasmaPower:
Header could be nil here, so we should check err first. Right now this results in a vague "method handler crashed" error message from the RPC if you eth_call against a header that doesn't exist, instead of the much better "header not found".


magicxyyz (approved):
LGTM

PR #314, opened 2024-05-02
fix pebble metrics registration [NIT-2468]

magicxyyz:
Pebble gauge metrics are stuck when twice opening same database (same metrics namespace). That happens in `openInitializeChainDb`: https://github.com/OffchainLabs/nitro/blob/e6afd28add5c9ff8be5952e3fe546d70015841af/cmd/nitro/init.go#L175-L181 

Gauge metrics are created with `metrics.NewRegisteredGauge` and it fails silently if metric name already exists. The Meter metrics work despite that `metrics.NewRegisteredMeter` is used, because it calls `metrics.GetOrRegisterMeter` internally.

This PR fixes the issue by using `metrics.GetOrRegister*` functions instead.

### DIFF
--- a/ethdb/pebble/pebble.go
+++ b/ethdb/pebble/pebble.go
@@ -233,19 +233,19 @@ func New(file string, cache int, handles int, namespace string, readonly bool, e
 	}
 	db.db = innerDB
 
-	db.compTimeMeter = metrics.NewRegisteredMeter(namespace+"compact/time", nil)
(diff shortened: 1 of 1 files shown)


Tristan-Wilson (approved):
LGTM

magicxyyz:
just added a PR to upstream the fix: https://github.com/ethereum/go-ethereum/pull/29699

PR #317, opened 2024-05-09
Separate Wasmstore from normal db: initial

tsahee:


arbitrum/recordingdb.go
```diff
@@ -57,10 +58,10 @@ func (db *RecordingKV) Get(key []byte) ([]byte, error) {
 		res, err = db.diskDb.Get(key)
 	} else if ok, _ := rawdb.IsActivatedAsmKey(key); ok {
 		// Arbitrum: the asm is non-consensus
-		return db.diskDb.Get(key)
+		return db.wasmDb.Get(key)
 	} else if ok, _ := rawdb.IsActivatedModuleKey(key); ok {
 		// Arbitrum: the module is non-consensus (only its hash is)
-		return db.diskDb.Get(key)
+		return db.wasmDb.Get(key)
```
PlasmaPower, line 64: Will these ever be called? Shouldn't they be called on the wasm database instead?

tsahee: moved some things around and now it is indeed not needed

core/state/journal_arbitrum.go
```diff
@@ -41,8 +41,11 @@ type EvictWasm struct {
 }
 
 func (ch EvictWasm) revert(s *StateDB) {
-	asm := s.GetActivatedAsm(ch.ModuleHash) // only happens in native mode
-	CacheWasmRust(asm, ch.ModuleHash, ch.Version, ch.Debug)
+	asm, err := s.TryGetActivatedAsm(ch.ModuleHash) // only happens in native mode
+	if err == nil && len(asm) == 0 {
```
PlasmaPower, line 45: I don't think this check is right. Maybe it should be `if err != nil || len(asm) == 0 {`?

core/state/journal_arbitrum.go
```diff
@@ -41,8 +41,11 @@ type EvictWasm struct {
...
+		//if we failed to get it - it's not in the current rust cache
+		CacheWasmRust(asm, ch.ModuleHash, ch.Version, ch.Debug)
```
PlasmaPower, line 47: I'm not really sure what's happening here. If we failed to get asm, why are we caching that same asm? Don't we need to regenerate it or something?

core/state/statedb_arbitrum.go
```diff
@@ -237,9 +234,13 @@ func (s *StateDB) StartRecording() {
 }
 
 func (s *StateDB) RecordProgram(moduleHash common.Hash) {
+	asm, err := s.TryGetActivatedAsm(moduleHash)
+	if err != nil {
+		log.Crit("can't find activated wasm while recording", "modulehash", moduleHash)
```
PlasmaPower, line 239: Is the program guaranteed to be in cache here? I don't think we can trust that it's in the database, e.g. if the tx reverts.

internal/ethapi/api.go
```diff
@@ -962,12 +962,17 @@ type OverrideAccount struct {
 // StateOverride is the collection of overridden accounts.
 type StateOverride map[common.Address]OverrideAccount
 
+var arbStorageAddress = common.HexToAddress("0xA4B05FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF")
```
PlasmaPower, line 965: There's already `types.ArbosStateAddress`

PlasmaPower (changes requested):

core/state/journal_arbitrum.go
tsahee: yeah, it's wrong. We shoul only cache it if we succeed.

core/state/statedb_arbitrum.go
tsahee: yes. It's called from arbos/programs/native.go:callProgram. Before we get to it we take the asm and module and store them at least to the statedb.

PlasmaPower (approved):
LGTM

PR #319, opened 2024-05-17
Stylus cache tag

tsahee:
support tag caches for wasm. Stored in ethdb and default to 0.


arbitrum/apibackend.go
```diff
@@ -101,8 +103,15 @@ func createRegisterAPIBackend(backend *Backend, filterConfig filters.Config, fal
 	if err != nil {
 		return nil, err
 	}
+	// discard stylus-tag on any call made from api database
+	dbForAPICalls := backend.chainDb
+	wasmStore, tag := backend.chainDb.WasmDataBase()
+	if tag != 0 {
+		dbForAPICalls = rawdb.WrapDatabaseWithWasm(backend.chainDb, wasmStore, tag)
```
PlasmaPower, line 110: I think the final argument here should be 0, not `tag`

PlasmaPower (approved):
LGTM

PR #321, opened 2024-05-23
Arbitrum APIBackend's GetTransaction should return false when tx is not found

ganeshvanahalli:
Resolves NIT-2544


joshuacolvin0 (approved):
LGTM

PR #326, opened 2024-06-10
Make arbitrum APIBackend's method stateAndHeaderFromHeader public. Used in rebuilding of wasm store

ganeshvanahalli:
Part of NIT-2503


magicxyyz (approved):
LGTM

PR #328, opened 2024-06-13
Improve pruning: performance, config and progress logging 

magicxyyz:
part of NIT-2541

This PR:
* reintroduces performance improvement previously removed while fixing upstream geth merge
```
// note: we are passing data.Root as stateRoot here, to skip the check for stateRoot existence in trie.newTrieReader,
// we already check that when opening state trie and reading the account node
trieID := trie.StorageTrieID(data.Root, key, data.Root)
storageTr, err := trie.NewStateTrie(trieID, sdb.TrieDB())
```

* enables cleans cache in triedb used for traversing storage tries in `dumpRawTrieDescendants` - when opening storage trie with `trie.NewStateTrie` (using the trick above), a node with hash equal `data.Root` is read twice from triedb (once in when getting a `trie.Reader`, then when reading the node). Caching the node should save some of KeyValueStore reads.

* adds `pruner.Config.Threads` option

* adds extra progress logging when a thread traversing storage trie for given account takes longer then 5 minutes

### DIFF
--- a/core/state/pruner/pruner.go
+++ b/core/state/pruner/pruner.go
@@ -385,7 +385,10 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
 				output.Put(data.CodeHash, nil)
 			}
 			if data.Root != (common.Hash{}) {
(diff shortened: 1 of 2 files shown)


PlasmaPower (approved):
LGTM

PR #335, opened 2024-07-09
[NIT-1168] Metrics for total gas used for eth_call and total gas used for eth_estimateGas

amsanghi:
used by https://github.com/OffchainLabs/nitro/pull/2478

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -54,8 +54,8 @@ import (
 
 var errBlobTxNotSupported = errors.New("signing blob transactions not supported")
 var (
-	gasUsedEthEstimateGasGauge = metrics.NewRegisteredGauge("api/eth_call/gas", nil)
-	gasUsedEthCallGauge        = metrics.NewRegisteredGauge("api/eth_estimateGas/gas", nil)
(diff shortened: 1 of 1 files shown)


internal/ethapi/api.go
```diff
@@ -52,6 +53,10 @@ import (
 )
 
 var errBlobTxNotSupported = errors.New("signing blob transactions not supported")
+var (
+	gasUsedEthEstimateGasGauge = metrics.NewRegisteredGauge("api/eth_call/gas", nil)
+	gasUsedEthCallGauge        = metrics.NewRegisteredGauge("api/eth_estimateGas/gas", nil)
```
joshuacolvin0, line 58: how about something like `rpc/gas_used/eth_call`?  Seems to fit in better to existing metrics.

Metric is only ever incremented, so should be a counter instead of a gauge.

joshuacolvin0 (changes requested):

internal/ethapi/api.go
amsanghi: Done.

Tristan-Wilson (approved):

joshuacolvin0 (approved):
LGTM

joshuacolvin0 (approved):
LGTM

PR #338, opened 2024-07-11
Switch merge checks action to pull_request_target event

PlasmaPower:
This is a copy of https://github.com/OffchainLabs/nitro/pull/2477 for the geth repo. This should allow the check to run on 3rd party PRs from forks. CI will need bypassed to merge this, as the "design approved check" won't appear, because pull_request_target only runs when it's on the base branch.


joshuacolvin0 (approved):
LGTM

PR #339, opened 2024-07-11
add support for wasm compile targets

magicxyyz:
part of NIT-2631

This PR: 
* changes wasm store schema to support storing assembly for multiple targets
* adds support for multiple targets in StateDB, including recording UserWasm for specific targets


core/state/database_arbitrum.go
```diff
@@ -7,16 +7,11 @@ import (
 	"github.com/ethereum/go-ethereum/core/rawdb"
 )
 
-func (db *cachingDB) ActivatedAsm(moduleHash common.Hash) ([]byte, error) {
+func (db *cachingDB) ActivatedAsm(targetName string, moduleHash common.Hash) ([]byte, error) {
 	if asm, _ := db.activatedAsmCache.Get(moduleHash); len(asm) > 0 {
```
magicxyyz, line 11: this cache is currently broken - the cache can be populated with asms for different targets

I'll either cache whole asmMap for given moduleHash OR only use cache when targetName == local target name.

core/state/statedb_arbitrum.go
```diff
@@ -233,16 +240,13 @@ func (s *StateDB) StartRecording() {
 	s.arbExtraData.userWasms = make(UserWasms)
 }
 
-func (s *StateDB) RecordProgram(moduleHash common.Hash) {
-	asm, err := s.TryGetActivatedAsm(moduleHash)
+func (s *StateDB) RecordProgram(targetNames []string, moduleHash common.Hash) {
```
tsahee, line 243: add here a small condition that if len(targetNames) ==0 -> do nothing and return.
Eventually, that's what we'll use to avoid unnecessary work when recording.

core/rawdb/accessors_state_arbitrum.go
```diff
@@ -22,15 +22,76 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
-// Stores the activated asm and module for a given codeHash
-func WriteActivation(db ethdb.KeyValueWriter, moduleHash common.Hash, asm, module []byte) {
-	key := ActivatedAsmKey(moduleHash)
+const (
+	TargetWavm = "wavm"
+	TargetArm  = "arm"
+	TargetX86  = "x86"
```
tsahee, line 28: not 100% certain - but I think maybe TargetArm and TargetAsm should be identical to the runtime.GOARCH on arm64/x86 architectures, which will be compatible with some exiting code

tsahee (commented):
very good.
tiny suggestions

core/rawdb/accessors_state_arbitrum.go
magicxyyz: I used GOARCH strings + I renamed the constants to TargetArm64 and TargetAmd64 to make it more consistent

core/rawdb/accessors_state_arbitrum.go
```diff
@@ -22,15 +22,76 @@ import (
...
+}
 
-	key = ActivatedModuleKey(moduleHash)
-	if err := db.Put(key[:], module); err != nil {
-		log.Crit("Failed to store activated wasm module", "err", err)
+// Retrieves wasm schema version, if the corresponding key is not found returns version 0
+func ReadWasmSchemaVersion(db ethdb.KeyValueReader) []byte {
+	version, err := db.Get(wasmSchemaVersionKey)
+	if err != nil || len(version) == 0 {
+		return []byte{0}
```
gligneul, line 94: Should we log the error if it is something other than not found?

core/state/statedb_arbitrum.go
```diff
@@ -76,38 +73,48 @@ func NewStylusPrefix(dictionary byte) []byte {
...
-	if err != nil {
-		s.setError(fmt.Errorf("failed to load module for %x: %v", moduleHash, err))
+	var err error
+	asmMap = make(map[string][]byte, len(targetNames))
+	for _, target := range targetNames {
+		asm, dbErr := s.db.ActivatedAsm(target, moduleHash)
+		if dbErr == nil {
+			asmMap[target] = asm
+		} else {
+			err = errors.Join(fmt.Errorf("failed to read activated asm from database for target %v and module %v: %w", target, moduleHash, dbErr))
```
gligneul, line 114:
```suggestion
			err = errors.Join(fmt.Errorf("failed to read activated asm from database for target %v and module %v: %w", target, moduleHash, dbErr), err)
```

gligneul (commented):
Looks good, just minor comments.

core/vm/interface.go
```diff
@@ -29,9 +29,9 @@ import (
 // StateDB is an EVM database for full state querying.
 type StateDB interface {
 	// Arbitrum: manage Stylus wasms
-	ActivateWasm(moduleHash common.Hash, asm, module []byte)
-	TryGetActivatedAsm(moduleHash common.Hash) (asm []byte, err error)
-	GetActivatedModule(moduleHash common.Hash) (module []byte)
+	ActivateWasm(moduleHash common.Hash, asm map[string][]byte)
```
ganeshvanahalli, line 32:
```suggestion
	ActivateWasm(moduleHash common.Hash, asmMap map[string][]byte)
```

core/rawdb/accessors_state_arbitrum.go
magicxyyz: I changed ReadWasmSchemaVersion to return error and added handling it on nitro side - we already have a regex there to check if the error is "not found" error (unfortunately, pebble and leveldb return different errors, maybe we shouold patch that in future?)

tsahee (approved):
LGTM

core/state/statedb_arbitrum.go
```diff
@@ -76,38 +74,48 @@ func NewStylusPrefix(dictionary byte) []byte {
...
-	if err != nil {
-		s.setError(fmt.Errorf("failed to load module for %x: %v", moduleHash, err))
+	var err error
+	asmMap = make(map[rawdb.Target][]byte, len(targets))
+	for _, target := range targets {
+		asm, dbErr := s.db.ActivatedAsm(target, moduleHash)
+		if dbErr == nil {
+			asmMap[target] = asm
+		} else {
+			err = errors.Join(fmt.Errorf("failed to read activated asm from database for target %v and module %v: %w", target, moduleHash, dbErr), err)
```
PlasmaPower, line 115: Side note, but I had no idea `errors.Join` existed, that's really helpful!

PlasmaPower (changes requested):

gligneul (approved):

PlasmaPower (commented):
LGTM (pending nitro PR approval)

PlasmaPower (approved):

gligneul (approved):

PR #345, opened 2024-07-18
Merge 1.14.0

amsanghi:
fixes: NIT-2673

Merge upstream geth's v1.14.0 release into our geth fork:https://github.com/ethereum/go-ethereum/releases/tag/v1.14.0

nitro PR- https://github.com/OffchainLabs/nitro/pull/2517

# Testing done
geth and nitro tests pass


internal/ethapi/transaction_args.go
```diff
@@ -474,25 +484,27 @@ func (args *TransactionArgs) ToMessage(globalGasCap uint64, header *types.Header
 		// ToMessage recurses once to allow ArbOS to intercept the result for all callers
 		// ArbOS uses this to modify globalGasCap so that the cap will ignore this tx's specific L1 data costs
 		core.InterceptRPCGasCap(&globalGasCap, msg, header, state)
-		return args.ToMessage(globalGasCap, header, nil, runMode) // we pass a nil to avoid another recursion
+		err := args.CallDefaults(globalGasCap, baseFee, chainID, gasLimitNotSetByUser)
+		// If we fail to call defaults, we should panic because it's a programming error since we've already called it once
+		if err != nil {
+			panic(fmt.Sprintf("CallDefaults failed: %v", err))
+		}
+		return args.ToMessage(baseFee, globalGasCap, header, nil, runMode, chainID, gasLimitNotSetByUser) // we pass a nil to avoid another recursion
```
tsahee, line 492: I'd like to hear more about that.
Also: @Tristan-Wilson

core/blockchain.go
```diff
@@ -785,15 +781,18 @@ func (bc *BlockChain) setHeadBeyondRoot(head uint64, time uint64, root common.Ha
 			// last step, however the direction of SetHead is from high
 			// to low, so it's safe to update in-memory markers directly.
 			bc.currentBlock.Store(newHeadBlock.Header())
-			headBlockGauge.Update(int64(newHeadBlock.NumberU64()))
+			headBlockGauge.Update(newHeadBlock.Number().Int64())
```
tsahee, line 784: It's not trivial - but we need to take changes we made to SetHeadBeyondRoot and apply them to rewindPathHead and rewindHashHead, calling the original rewindHead (from v1.14.0) from setHeadBeyondRoot

tsahee (changes requested):
comment on blockchain.go
question about transaction_arg.go

internal/ethapi/transaction_args.go
amsanghi: Earlier there was no concept of CallDefaults before calling ToMessage, but now there is.
So when we recurse in ToMessage for 

> // Arbitrum: raise the gas cap to ignore L1 costs so that it's compute-only
	if state != nil {

We need to call CallDefaults to set the new globalGasCap (updated by InterceptRPCGasCap) again.
But because the CallDefaults was already called earlier the the args.Gas is already set to a non zero value (even though this value was not given by the user).
So to allow us to update args.Gas even to though it is set, i introduced a new input to function CallDefaults called gasLimitNotSetByUser which will help in determining whether we should be updating args.Gas even if it holds a non zero value (cause this non zero value was not set by the user).

eljobe (approved):
LGTM

tsahee (changes requested):

core/state/state_object.go
```diff
@@ -359,15 +358,12 @@ func (s *stateObject) updateTrie() (Trie, error) {
...
 	}
-	if len(keysToDelete) > 0 {
-		sort.Slice(keysToDelete, func(i, j int) bool { return bytes.Compare(keysToDelete[i][:], keysToDelete[j][:]) < 0 })
-		for _, key := range keysToDelete {
-			if err := tr.DeleteStorage(s.address, key[:]); err != nil {
-				s.db.setError(err)
-				return nil, err
-			}
-			s.db.StorageDeleted += 1
+	for _, key := range deletions {
```
tsahee, line 361: This changes the order in which keys are deleted compared to previous version, which could is very important because the order affects record/replay (as it is, some blocks recording won't fit the replay).
The behavior only matters if s.deterministic is true, but in this case it should be consistent with the previous version.
"deletions" here seems to be identical to "keysToDelete" in the previous deterministic case.
If I'm right, all we need to do is check s.deterministic, and if it's true, before deleting keys we need to
```
sort.Slice(keysToDelete, func(i, j int) bool { return bytes.Compare(keysToDelete[i][:], keysToDelete[j][:]) < 0 })
```
and that will be consistent with previous behaviour.

tsahee (changes requested):

tsahee (changes requested):

metrics/sample_test.go
```diff
@@ -87,25 +87,6 @@ func BenchmarkUniformSample1028(b *testing.B) {
...
-	benchmarkSample(b, NewSlidingTimeWindowArraySample(257))
-}
-
-func BenchmarkSlidingWindowArraySample514(b *testing.B) {
-	benchmarkSample(b, NewSlidingTimeWindowArraySample(514))
-}
-
-func BenchmarkSlidingWindowArraySample1028(b *testing.B) {
-	benchmarkSample(b, NewSlidingTimeWindowArraySample(1028))
-}
```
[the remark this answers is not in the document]
amsanghi: Done.

core/state/state_object.go
amsanghi: Done.

core/blockchain.go
```diff
@@ -675,6 +672,221 @@ func (bc *BlockChain) SetSafe(header *types.Header) {
...
+			lastFullBlock = head.Number.Uint64()
+			lastFullBlockHash = head.Hash()
+		}
+		log.Debug("Skipping block with threshold state", "number", head.Number, "hash", head.Hash(), "root", head.Root)
+		head = bc.GetHeader(head.ParentHash, head.Number.Uint64()-1) // Keep rewinding
+	}
+}
+
+// rewindPathHead implements the logic of rewindHead in the context of path scheme.
+func (bc *BlockChain) rewindPathHead(head *types.Header, root common.Hash, rewindLimit uint64) (*types.Header, uint64, bool) {
```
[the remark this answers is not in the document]
amsanghi: Done.

core/blockchain.go
```diff
@@ -159,13 +157,16 @@ type CacheConfig struct {
 }
 
 // arbitrum: exposing CacheConfig.triedbConfig to be used by Nitro when initializing arbos in database
-func (c *CacheConfig) TriedbConfig() *triedb.Config {
-	return c.triedbConfig()
+func (c *CacheConfig) TriedbConfig(isVerkle bool) *triedb.Config {
+	return c.triedbConfig(isVerkle)
```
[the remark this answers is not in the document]
amsanghi: Done.

tsahee (commented):
LGTM.
seems like nitro still has CI problems, waiting for that before merging

tsahee (approved):
LGTM

PR #347, opened 2024-08-05
Add a notion of zombies to preserve empty account behavior

PlasmaPower:
Points into https://github.com/OffchainLabs/go-ethereum/pull/346

Geth statedb changes in v1.14.2 mean that the old behavior of SubBalance amount 0 in respect to empty accounts has changed. This PR adds back in the old behavior using a notion of zombies.

Requires https://github.com/OffchainLabs/nitro/pull/2732 for the new method to be used on the nitro side


amsanghi (approved):
LGTM, thanks for fixing this!

PR #350, opened 2024-08-09
Add config option to change HTTPBodyLimit and WSReadLimit

amsanghi:
Fixes: NIT-2725

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -110,6 +110,10 @@ func New(conf *Config) (*Node, error) {
 	}
 	server := rpc.NewServer()
 	server.SetBatchLimits(conf.BatchRequestLimit, conf.BatchResponseMaxSize)
+	// If a custom http body limit is set, apply it to the server.
+	if conf.HTTPBodyLimit != 0 {
+		server.SetHTTPBodyLimit(conf.HTTPBodyLimit)
+	}
 	node := &Node{
 		config:        conf,
 		inprocHandler: server,
--- a/node/node.go
+++ b/node/node.go
@@ -423,6 +423,12 @@ func (n *Node) startRPC() error {
 		batchResponseSizeLimit: n.config.BatchResponseMaxSize,
 		apiFilter:              n.apiFilter,
 	}
+	if n.config.HTTPBodyLimit != 0 {
+		rpcConfig.httpBodyLimit = n.config.HTTPBodyLimit
+	}
+	if n.config.WSReadLimit != 0 {
+		rpcConfig.wsReadLimit = n.config.WSReadLimit
+	}
 
 	initHttp := func(server *httpServer, port int) error {
 		if err := server.setListenAddr(n.config.HTTPHost, port); err != nil {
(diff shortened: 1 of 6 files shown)


node/node.go
```diff
@@ -110,6 +110,10 @@ func New(conf *Config) (*Node, error) {
 	}
 	server := rpc.NewServer()
 	server.SetBatchLimits(conf.BatchRequestLimit, conf.BatchResponseMaxSize)
+	// If a custom http body limit is set, apply it to the server.
+	if conf.HTTPBodyLimit != 0 {
+		server.SetHTTPBodyLimit(conf.HTTPBodyLimit)
```
PlasmaPower, line 115: I'm not sure this is the right server / the right place to set it. Have you verified with a nitro-val instance that it increases the max request size for the authenticated websocket port?

amsanghi: The type of the error mentioned in the linear ticket is only available in this server https://github.com/OffchainLabs/go-ethereum/blob/master/rpc/http.go#L345 "content length too large"

Though happy to test this out as well, can you please help me with the steps, i am not sure how to send the request to test this.

PlasmaPower: I meant another instance of the same server type. To test this, first startup nitro-val:

```
make target/bin/nitro-val
target/bin/nitro-val --auth.addr 127.0.0.1 --auth.port 52000 --auth.jwtsecret <(echo 71776572747975696f706173646667686a6b6c7a786376626e6d313233343536) --log-level trace
```

Then, go get a jwt from http://jwtbuilder.jamiekurtz.com/ . I've setup the above secret so it's the default key there. The jwt only lasts for 60 seconds, so if you take too long you might get "stale token". You can fix that by hitting the "now" button to the right of the "Issued At" field on that page and then regenerating the token with "create signed jwt" at the bottom (then "copy jwt to clipboard").

Finally, try curling this, with nitro-val still running:
```
curl http://127.0.0.1:52000 -H 'Content-Type: application/json' -H 'Authorization: Bearer YOUR_JWT_TOKEN_HERE' -H 'Content-Length: 1000000000'
```

If successful, it should hang (waiting for input). However, on your branch, I currently get `content length too large (1000000000>134217728)`, indicating that the request size cap hasn't been successfully changed.

We should also make sure that the websocket limit is raised. To test that, first install wscat with `npm install -g wscat`. With that installed, try sending a large websocket request (with a JWT again, make sure it's fresh or you'll get a nondescript 401):

```
(cat /dev/zero | tr '\0' 'a' | head -c 150000000; echo; sleep 1) | wscat -c ws://127.0.0.1:52000 -H 'Content-Type: application/json' -H 'Authorization: Bearer YOUR_JWT_TOKEN_HERE' | cat
```

The output of this doesn't change depending on if the limit was hit, but if you look at the nitro-val logs, it'll say `invalid character 'a' looking for beginning of value` if the request successfully got past the limit or `websocket: read limit exceeded` if it hit the limit.

amsanghi: Tested, both the conditions are being satisfied now!

PlasmaPower (approved):
LGTM

PR #352, opened 2024-08-22
Integrate CaptureArbitrumStorageGet/Set into the prestate tracer

ganeshvanahalli:


eth/tracers/native/prestate.go
```diff
@@ -42,15 +42,18 @@ func init() {
...
 type account struct {
-	Balance *big.Int                    `json:"balance,omitempty"`
-	Code    []byte                      `json:"code,omitempty"`
-	Nonce   uint64                      `json:"nonce,omitempty"`
-	Storage map[common.Hash]common.Hash `json:"storage,omitempty"`
-	empty   bool
+	Balance         *big.Int                    `json:"balance,omitempty"`
+	Code            []byte                      `json:"code,omitempty"`
+	Nonce           uint64                      `json:"nonce,omitempty"`
+	Storage         map[common.Hash]common.Hash `json:"storage,omitempty"`
```
diegoximenes, line 48: How about adding an empty line after this one?
All lines above exist in the original geth, so adding an empty line here would avoid changing those lines, which can simplify the process of updating this fork with changes from the original geth in the future.

diegoximenes (changes requested):

eth/tracers/native/prestate.go
```diff
@@ -279,9 +305,22 @@ func (t *prestateTracer) lookupAccount(addr common.Address) {
 // lookupStorage fetches the requested storage slot and adds
 // it to the prestate of the given contract. It assumes `lookupAccount`
 // has been performed on the contract before.
-func (t *prestateTracer) lookupStorage(addr common.Address, key common.Hash) {
+func (t *prestateTracer) lookupStorage(addr common.Address, key, mappedKey common.Hash, isArbitrumStorage bool) {
```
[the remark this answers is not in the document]
ganeshvanahalli: Good point! I agree

eth/tracers/native/prestate.go
ganeshvanahalli: I agree

diegoximenes (approved):

diegoximenes (changes requested):

eth/tracers/native/prestate.go
```diff
@@ -208,7 +214,10 @@ func (t *prestateTracer) processDiffState() {
 			continue
 		}
 		modified := false
-		postAccount := &account{Storage: make(map[common.Hash]common.Hash)}
+		postAccount := &account{
+			Storage:         make(map[common.Hash]common.Hash),
+			ArbitrumStorage: make(map[common.Hash]common.Hash),
+		}
```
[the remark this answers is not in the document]
ganeshvanahalli: makes sense!

eth/tracers/native/prestate.go
```diff
@@ -46,11 +46,15 @@ type account struct {
 	Code    []byte                      `json:"code,omitempty"`
 	Nonce   uint64                      `json:"nonce,omitempty"`
 	Storage map[common.Hash]common.Hash `json:"storage,omitempty"`
-	empty   bool
+
+	ArbitrumStorage map[common.Hash]common.Hash `json:"arbitrumStorage,omitempty"`
+
+	empty            bool
```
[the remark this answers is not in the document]
ganeshvanahalli: nice suggestion! fixed it

diegoximenes (approved):

core/tracing/hooks.go
```diff
@@ -148,8 +148,8 @@ type (
 	LogHook = func(log *types.Log)
 
 	CaptureArbitrumTransferHook   = func(from, to *common.Address, value *big.Int, before bool, purpose string)
-	CaptureArbitrumStorageGetHook = func(key common.Hash, depth int, before bool)
-	CaptureArbitrumStorageSetHook = func(key, value common.Hash, depth int, before bool)
+	CaptureArbitrumStorageGetHook = func(addr common.Address, key, mappedKey common.Hash, depth int, before bool)
+	CaptureArbitrumStorageSetHook = func(addr common.Address, key, mappedKey, value common.Hash, depth int, before bool)
```
tsahee, line 152: addr should always be the arbosStorage address. We can add it here but not really necessary.
We shouldn't need mapped key because that's what "key" should mean. Storage is just key-value pairs and we don't care about the non-mapped key.

eth/tracers/native/prestate.go
```diff
@@ -47,10 +47,13 @@ type account struct {
 	Nonce   uint64                      `json:"nonce,omitempty"`
 	Storage map[common.Hash]common.Hash `json:"storage,omitempty"`
 	empty   bool
+
+	ArbitrumStorage  map[common.Hash]common.Hash `json:"arbitrumStorage,omitempty"`
+	arbStorageKeyMap map[common.Hash]common.Hash
```
tsahee, line 52: Prestate is just about finding all the places in storage that chaged.
We're not trying to add new things there. It already supports writes that happen during EVM by mimicing an SLOAD/SSTORE.
We just want to add the accesses that happened before EVM execution to the same place.
So we shouldn't need new entries. It should all be normal entries in the arbosstorage account.

tsahee (dismissed):

joshuacolvin0 (approved):
LGTM

PR #356, opened 2024-09-10
add arbos v32 constant as StylusActivationFix

tsahee:


tsahee:
replaced with https://github.com/OffchainLabs/go-ethereum/pull/360

PR #359, opened 2024-09-12
Bubble up errors reading snapshot info from db

PlasmaPower:
This can otherwise cause database corruption and invalid state due to database errors as the statedb will think that the account or storage simply doesn't exist.


joshuacolvin0 (approved):
LGTM

PR #362, opened 2024-09-26
fix MessageRunMode usage

magicxyyz:
Part of NIT-2812

This PR:

* adds MessageRunMode parameter to ApplyTransactionWithResultFilter, so it can be called with different modes

* changes mode used in `StateProcessor.Process` from `MessageCommitMode` to `MessageReplayMode`


tsahee (approved):
LGTM

PR #367, opened 2024-10-29
Re-add kilic BLS12-381 library

PlasmaPower:
Re-adds the in-tree kilic BLS12-381 library removed by upstream geth in https://github.com/ethereum/go-ethereum/pull/29441

Unfortunately we need to keep it around for now for nitro

Points into https://github.com/OffchainLabs/go-ethereum/pull/345


joshuacolvin0 (approved):
LGTM

PR #369, opened 2024-10-29
Add a boolean field to tx receipt object indicating if the tx was timeboosted

ganeshvanahalli:
This PR introduces a new field to the receipts object/s returned by `eth_getTransactionReceipt` and `eth_getBlockReceipts`, indicating whether a transaction was timeboosted or not. This field is a boolean named as `timeboosted`.

The PR is based of of `wasm-store-compile-target` branch (revived) instead of master because the current main Timeboost PR's geth pin is pointing to this branch and we do this to be in sync with the main PR.

Corresponding nitro PR- https://github.com/OffchainLabs/nitro/pull/2760
Resolves NIT-2836


internal/ethapi/api.go
```diff
@@ -1963,6 +1963,14 @@ func marshalReceipt(ctx context.Context, receipt *types.Receipt, blockHash commo
...
 			}
 		}
+
+		blockMetadata, err := backend.BlockMetadataByNumber(blockNumber)
+		if err != nil {
+			return nil, err
+		}
+		if blockMetadata != nil && blockMetadata.IsTxTimeboosted(txIndex) {
+			fields["timeboosted"] = true
+		}
```
Tristan-Wilson, line 1973: Should we put timeboosted = false for Blocks with metadata but Txs that are not timeboosted?

Tristan-Wilson (changes requested):

internal/ethapi/api.go
ganeshvanahalli: Nice catch, fixed it.

Tristan-Wilson (dismissed):

common/types.go
```diff
@@ -486,3 +486,20 @@ func (b PrettyBytes) TerminalString() string {
...
+		return false, errors.New("blockMetadata is not set")
+	}
+	if txIndex < 0 {
+		return false, fmt.Errorf("invalid transaction index- %d, should be positive", txIndex)
+	}
+	maxTxCount := (len(b) - 1) * 8
+	if txIndex >= maxTxCount {
+		return false, nil
+	}
+	return b[1+(txIndex/8)]&(1<<(txIndex%8)) != 0, nil
```
eljobe, line 504: This line needs a comment.
8 seems like a magical number and it isn't clear how the bitwise & is checking if the transaction was timeboosted or not.

eljobe (changes requested):

common/types.go
ganeshvanahalli: I agree. Currently the explanation for this exists on the sequencer side where blockMetadata originates https://github.com/OffchainLabs/nitro/blob/071aa54de1a8c2f510b5d11835bad9124eb38f37/execution/gethexec/executionengine.go#L615-L619 but having it here as well is a good idea

tsahee (approved):
LGTM

PR #304, opened 2024-04-18
Merge v1.13.6

ganeshvanahalli:
Merge upstream geth's v1.13.6 release into our geth fork:https://github.com/ethereum/go-ethereum/releases/tag/v1.13.6 
* Major changes are 
    * optimization and refactoring of gas estimation
    * moving logging over from custom in-house logger (fork of [log15](https://github.com/inconshreveable/log15)) to [slog](https://go.dev/blog/slog)
* On M2 mac, metrics/influxdb/ tests fail due to different floating point results

_Uses Tristian's https://github.com/Tristan-Wilson/MyScripts/blob/main/bin/merge-helper_

nitro PR- https://github.com/OffchainLabs/nitro/pull/2237
Part of NIT-2394

# Testing done
geth and nitro tests pass

PR #373, opened 2024-11-08
replace MessageRunMode with a struct

magicxyyz:
This PR:
* replaces `MessageRunMode` enum with `MessageRunContext` struct, that apart from run mode now contains wasm cache tag and wasm targets list
* Refactors passing of wasm targets and wasm cache tag, so they are no longer wired through `cachingDB` and `ethdb.Database`

Resolves NIT-2827


cmd/evm/internal/t8ntool/execution.go
```diff
@@ -213,7 +213,7 @@ func (pre *Prestate) Apply(vmConfig vm.Config, chainConfig *params.ChainConfig,
 			rejectedTxs = append(rejectedTxs, &rejectedTx{i, errMsg})
 			continue
 		}
-		msg, err := core.TransactionToMessage(tx, signer, pre.Env.BaseFee, core.MessageCommitMode)
+		msg, err := core.TransactionToMessage(tx, signer, pre.Env.BaseFee, core.NewMessageCommitContext(nil)) // TODO: do we need to specify wasm targets here if chainConfig.IsArbitrum == false?
```
magicxyyz, line 216: do we need to specify wasm targets here if chainConfig.IsArbitrum == false?

tsahee: per other comments, I think we should allow init with nil to mean localTarget
Also, if stylus isn't active, it should't be a problem to really have nil there

tsahee (changes requested):

internal/ethapi/api.go
```diff
@@ -1167,19 +1167,19 @@ func doCall(ctx context.Context, b Backend, args TransactionArgs, state *state.S
 	}
 
 	// Arbitrum: a tx can schedule another (see retryables)
-	result, err = runScheduledTxes(ctx, b, state, header, blockCtx, core.MessageGasEstimationMode, result)
+	result, err = runScheduledTxes(ctx, b, state, header, blockCtx, core.NewMessageGasEstimationContext(), result) // TODO: why do we use GasEstimationMode here?
```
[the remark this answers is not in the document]
magicxyyz: `runScheduledTxes` was added in: https://github.com/OffchainLabs/go-ethereum/pull/304/commits/018bd54e2ecdf494dce8f59e29cc083af9bdd74c#

it addressed following comment: https://github.com/OffchainLabs/go-ethereum/pull/304#discussion_r1571579419

that is how it looked like before: https://github.com/OffchainLabs/go-ethereum/blob/9830725715f92cd4ed1809b3d069af2ef25ae6e6/internal/ethapi/api.go#L1131-L1170

core/state_transition.go
```diff
@@ -160,24 +161,88 @@ type Message struct {
...
 )
 
+type MessageRunContext struct {
+	runMode messageRunMode
+
+	wasmCacheTag uint32
+	wasmTargets  []rawdb.WasmTarget
+}
+
+func NewMessageCommitContext(wasmTargets []rawdb.WasmTarget) *MessageRunContext {
```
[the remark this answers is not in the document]
magicxyyz: went with the second option, if len(wasmTargets) == 0 then wasmTargets = [localTarget]

eljobe:
Assigning to @magicxyyz to clean up the conflicts and resolve review feedback.

tsahee:
I want this to wait until after we cut the Arbos-40 WasmModuleRoot

core/state_transition.go
```diff
@@ -178,24 +179,112 @@ type Message struct {
...
+		wasmTargets = []rawdb.WasmTarget{rawdb.LocalTarget()}
+	}
+	return &MessageRunContext{
+		runMode:      messageCommitMode,
+		wasmCacheTag: 1,
+		wasmTargets:  wasmTargets,
+	}
+}
+
+func NewMessageReplayContext(wasmTargets []rawdb.WasmTarget) *MessageRunContext {
```
magicxyyz, line 209: * Replay and Prefetch shouldn't take targets

* add RecordContext that takes targets

core/state_processor.go
```diff
@@ -217,11 +219,11 @@ func MakeReceipt(evm *vm.EVM, result *ExecutionResult, statedb *state.StateDB, b
...
 // indicating the block was invalid.
 func ApplyTransaction(evm *vm.EVM, gp *GasPool, statedb *state.StateDB, header *types.Header, tx *types.Transaction, usedGas *uint64) (*types.Receipt, *ExecutionResult, error) {
-	return ApplyTransactionWithResultFilter(evm, gp, statedb, header, tx, usedGas, MessageReplayMode, nil)
+	return ApplyTransactionWithResultFilter(evm, gp, statedb, header, tx, usedGas, NewMessageReplayContext([]rawdb.WasmTarget{rawdb.LocalTarget()}), nil)
 }
 
-func ApplyTransactionWithResultFilter(evm *vm.EVM, gp *GasPool, statedb *state.StateDB, header *types.Header, tx *types.Transaction, usedGas *uint64, runMode MessageRunMode, resultFilter func(*ExecutionResult) error) (*types.Receipt, *ExecutionResult, error) {
-	msg, err := TransactionToMessage(tx, types.MakeSigner(evm.ChainConfig(), header.Number, header.Time, evm.Context.ArbOSVersion), header.BaseFee, runMode)
+func ApplyTransactionWithResultFilter(evm *vm.EVM, gp *GasPool, statedb *state.StateDB, header *types.Header, tx *types.Transaction, usedGas *uint64, runCtx *MessageRunContext, resultFilter func(*ExecutionResult) error) (*types.Receipt, *ExecutionResult, error) {
```
magicxyyz, line 225: can we put resultFilter into `MessageRunContext`?

core/state_transition.go
magicxyyz: only RecordContext and CommitContext should take more targets

magicxyyz: done, added MessageRecordingContext and removed targets params from Replay and Prefetch contexts constructors

tsahee (approved):
another small suggestion for later

eth/gasestimator/gasestimator.go
```diff
@@ -270,7 +270,7 @@ func run(ctx context.Context, call *core.Message, opts *Options) (*core.Executio
 	}
 
 	// Arbitrum: a tx can schedule another (see retryables)
-	result, err = opts.RunScheduledTxes(ctx, opts.Backend, dirtyState, opts.Header, evmContext, core.MessageGasEstimationMode, result)
+	result, err = opts.RunScheduledTxes(ctx, opts.Backend, dirtyState, opts.Header, evmContext, core.NewMessageGasEstimationContext(), result)
```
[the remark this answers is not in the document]
magicxyyz: good call, fixed.
Not passing the same instance of run context could break something in future (e.g. if we added some extra fields to run context the values would be lost there).

PR #376, opened 2024-11-19
add SyncMode option to PebbleExtraOptions

magicxyyz:
Resolves NIT-2952


diegoximenes (approved):

tsahee (approved):
LGTM

tsahee:
Need to spend some time thinking what should be default for us

ethdb/pebble/pebble.go
```diff
@@ -238,7 +238,7 @@ func New(file string, cache int, handles int, namespace string, readonly bool, e
 		fn:           file,
 		log:          logger,
 		quitChan:     make(chan chan error),
-		writeOptions: &pebble.WriteOptions{Sync: !ephemeral},
```
tsahee, line 241: I think I'll feel better if our default will stay "on" and we could later think if the logic that goes for geth turning it off also fits us

tsahee (changes requested):

tsahee (approved):
LGTM

PR #377, opened 2024-11-22
Filter transaction

ganeshvanahalli:
This PR enables filtering of malicious transactions and blocks via a `stateDB` object

Corresponding Nitro PR- https://github.com/OffchainLabs/nitro/pull/2807
Part of NIT-2848

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -19,6 +19,7 @@ package state
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"maps"
 	"math/big"
@@ -86,6 +87,7 @@ func (m *mutation) isDelete() bool {
 // commit states.
 type StateDB struct {
 	arbExtraData *ArbitrumExtraData // must be a pointer - can't be a part of StateDB allocation, otherwise its finalizer might not get called
+	arbTxFilter  int
 
 	db         Database
 	prefetcher *triePrefetcher
@@ -217,6 +219,19 @@ func New(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB, error) 
 		sdb.snap = sdb.snaps.Snapshot(root)
 	}
 	return sdb, nil
+}
+
+func (s *StateDB) FilterTx(withBlock bool) {
+	if s.arbTxFilter == 0 {
+		s.arbTxFilter = 1
+		if withBlock {
+			s.arbTxFilter = 2
+		}
+	}
+}
+
+func (s *StateDB) IsTxValid() bool {
+	return s.arbTxFilter == 1
 }
 
 // SetLogger sets the logger for account update hooks.
@@ -835,6 +850,9 @@ func (s *StateDB) RevertToSnapshot(revid int) {
 	revision := s.validRevisions[idx]
 	snapshot := revision.journalIndex
 	s.arbExtraData.unexpectedBalanceDelta = new(big.Int).Set(revision.unexpectedBalanceDelta)
+	if s.arbTxFilter == 1 {
+		s.arbTxFilter = 0
+	}
 
 	// Replay the journal to undo changes and remove invalidated snapshots
 	s.journal.revert(s, snapshot)
@@ -1220,6 +1238,9 @@ func (s *StateDB) GetTrie() Trie {
 // The associated block number of the state transition is also provided
 // for more chain context.
 func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, error) {
+	if s.arbTxFilter == 2 {
+		return common.Hash{}, errors.New("internal error")
+	}
 	// Short circuit in case any database failure occurred earlier.
 	if s.dbErr != nil {
 		return common.Hash{}, fmt.Errorf("commit aborted due to earlier error: %v", s.dbErr)
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -77,7 +77,8 @@ func (m *mutation) isDelete() bool {
 type arbFiltered int
 
 const (
-	txFiltered arbFiltered = 1 + iota
+	unFiltered arbFiltered = iota
+	txFiltered
 	blockFiltered
 )
 
@@ -231,7 +232,7 @@ func New(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB, error) 
 }
 
 func (s *StateDB) FilterTx(withBlock bool) {
-	if s.arbTxFilter == 0 {
+	if s.arbTxFilter == unFiltered {
 		s.arbTxFilter = txFiltered
 		if withBlock {
 			s.arbTxFilter = blockFiltered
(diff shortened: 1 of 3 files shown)


core/state/statedb.go
```diff
@@ -86,6 +87,7 @@ func (m *mutation) isDelete() bool {
 // commit states.
 type StateDB struct {
 	arbExtraData *ArbitrumExtraData // must be a pointer - can't be a part of StateDB allocation, otherwise its finalizer might not get called
+	arbTxFilter  int
```
eljobe, line 90: Maybe we could use an enum instead of an int so that each case could have a clear human-readable name in the code?

core/vm/interface.go
```diff
@@ -48,6 +48,10 @@ type StateDB interface {
 	// Arbitrum: preserve old empty account behavior
 	CreateZombieIfDeleted(common.Address)
 
+	// Arbitrum
```
eljobe, line 51: I'm surprised by this API. I would have expected that the api, would be something like, "Prevent transactions that have this (From and Nonce or some other unique transaction identifier) from being processed." But, this seems like it's just flipping some bit in the StateDB telling it to reject the "next" transaction or "next block." It would be great to have a few tests in this PR to demonstrate how callers can use this API to accomplish the desired filtering.

eljobe (changes requested):
Can you also update the PR comment to explain what's going on and why? The link to the Linear ticket is not sufficient for people who want to contribute to the project but don't have access to Linear.

core/state/statedb.go
```diff
@@ -74,6 +74,15 @@ func (m *mutation) isDelete() bool {
 	return m.typ == deletion
 }
 
+type arbFiltered int
+
+const (
+	txFiltered arbFiltered = 1 + iota
```
eljobe, line 80: I think there should be another value in the enum for the zero value.
Maybe `unfiltered` or `notFiltered`

ganeshvanahalli: added it

core/vm/interface.go
ganeshvanahalli: added a nitro system_test in this [PR](https://github.com/OffchainLabs/nitro/pull/2807)

core/state/statedb.go
```diff
@@ -73,6 +74,16 @@ func (m *mutation) isDelete() bool {
 	return m.typ == deletion
 }
 
+type arbFiltered int
+
+const (
+	unFiltered arbFiltered = iota
```
eljobe, line 80: Sorry to be so picky, but since unfiltered is a single word in English. I'd either use `unfiltered` or `notFiltered` here.

eljobe (dismissed):

core/state/statedb.go
```diff
@@ -1220,6 +1248,9 @@ func (s *StateDB) GetTrie() Trie {
 // The associated block number of the state transition is also provided
 // for more chain context.
 func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, error) {
+	if s.arbTxFilter == blockFiltered {
```
tsahee, line 1251: what about txFiltered?
I think this should also return an error as we expect filtered transactions to be reverted using RevertToSnapshot

core/state/statedb.go
```diff
@@ -835,6 +860,9 @@ func (s *StateDB) RevertToSnapshot(revid int) {
 	revision := s.validRevisions[idx]
 	snapshot := revision.journalIndex
 	s.arbExtraData.unexpectedBalanceDelta = new(big.Int).Set(revision.unexpectedBalanceDelta)
+	if s.arbTxFilter == txFiltered {
+		s.arbTxFilter = notFiltered
+	}
```
tsahee, line 865: there seems to be an assumption here that revert will always follow the last tx if it was filtered. This might be true, but should at least be checked - at least, check that the tx revert happened after that snapshot and/or fail taking a snapshot if we already have a reverted transaction.. something like that

core/state/statedb.go
```diff
@@ -219,6 +231,19 @@ func New(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB, error)
...
+func (s *StateDB) FilterTx(withBlock bool) {
+	if s.arbTxFilter == notFiltered {
+		s.arbTxFilter = txFiltered
+		if withBlock {
+			s.arbTxFilter = blockFiltered
+		}
+	}
+}
+
+func (s *StateDB) IsTxInvalid() bool {
```
tsahee, line 243: IsTxFiltered to keep naming consistent?

tsahee (dismissed):

core/state/statedb.go
ganeshvanahalli: txFiltered shouldn't be hit if managed correctly (i.e revert snapshot is called promptly after a failure), but its safer to error out anyway.

core/state/statedb.go
ganeshvanahalli: thanks for the suggestion, it does make sense to have an additional check in place to make sure revert snapshot is called in case of errors

ganeshvanahalli: I've addressed this by adding a panic

ganeshvanahalli: Modified the code in block_processor to include the above check. https://github.com/OffchainLabs/nitro/pull/2807/commits/b2618c992ae8a76500c7789c00efcea106c743be#diff-6e42525ac0041b38964449b89dc7af9261e780bb800266ec13b63e5c681004e4R300-R315

PlasmaPower (changes requested):

PlasmaPower:
Also, I think this flag might need to survive reverts

core/state/statedb.go
```diff
@@ -219,6 +224,16 @@ func New(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB, error)
 	return sdb, nil
 }
 
+func (s *StateDB) FilterTx() {
+	if !s.arbTxFilter {
+		s.arbTxFilter = true
```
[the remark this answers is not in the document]
ganeshvanahalli: yep, fixed it

core/state/statedb.go
```diff
@@ -86,6 +90,7 @@ func (m *mutation) isDelete() bool {
 // commit states.
 type StateDB struct {
 	arbExtraData *ArbitrumExtraData // must be a pointer - can't be a part of StateDB allocation, otherwise its finalizer might not get called
+	arbTxFilter  bool
```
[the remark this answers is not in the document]
ganeshvanahalli: moved it to arbExtraData for clarity, this also avoids merge conflicts in the future

ganeshvanahalli:
> Also, I think this flag might need to survive reverts

This is handled by persisting the flag in revisions

PlasmaPower:
I'm not sure, wouldn't that mean the flag could get reset to false if a contract reverts?

ganeshvanahalli:
> I'm not sure, wouldn't that mean the flag could get reset to false if a contract reverts?

if a contract reverts such that it leads to a tx failure, then we dont have to worry about the boolean as statedb would be reset to previous snapshot in the block_processor anyway. and if the contract revert doesn't cause the tx failure then geth would've taken a statedb snapshot before such an execution right? so that it can revert back to it? in which can revisions would persist the boolean's value

eljobe (dismissed):
LGMT

eljobe (approved):
LGTM

PR #381, opened 2024-12-09
Merge v1.14.4

amsanghi:
fixes: NIT-2678

Merge upstream geth's v1.14.4 release into our geth fork:https://github.com/ethereum/go-ethereum/releases/tag/v1.14.4

nitro PR- https://github.com/OffchainLabs/nitro/pull/2820

# Testing done
geth and nitro tests pass


core/state/statedb.go
```diff
@@ -1293,128 +1284,128 @@ func (s *StateDB) Commit(block uint64, deleteEmptyObjects bool) (common.Hash, er
...
+		}
+		if err := batch.Write(); err != nil {
+			return nil, err
+		}
+	}
+
+	if db := s.db.WasmStore(); db != nil && len(s.arbExtraData.activatedWasms) > 0 {
+		batch := db.NewBatch()
+		// Arbitrum: write Stylus programs to disk
+		for moduleHash, asmMap := range s.arbExtraData.activatedWasms {
```
tsahee, line 1353: I think we should follow geth's lead and add the activatedWasms to the stateUpdate struct (in newStateUpdate).
Not 100% certain - please look in forward geth versions to see if it makes sense

core/state/statedb.go
```diff
@@ -1135,94 +1117,74 @@ func (s *StateDB) deleteStorage(addr common.Address, addrHash common.Hash, root
...
 // In case (d), **original** account along with its storages should be deleted,
 // with their values be tracked as original value.
-func (s *StateDB) handleDestruction(nodes *trienode.MergedNodeSet) error {
-	// Short circuit if geth is running with hash mode. This procedure can consume
-	// considerable time and storage deletion isn't supported in hash mode, thus
-	// preemptively avoiding unnecessary expenses.
-	if s.db.TrieDB().Scheme() == rawdb.HashScheme {
-		return nil
-	}
+func (s *StateDB) handleDestruction() (map[common.Hash]*accountDelete, []*trienode.NodeSet, error) {
```
tsahee, line 1137: I'm not 100% certain I understand this part.
However, I think that it should be safe and preferrable to check if s.deterministic is on and in this case return nil, nil
Deleting storage was previously not supported on hashdb at all, so I think we can safely skip it on record/replay.
@PlasmaPower WDYT?

tsahee (dismissed):

core/state/statedb.go
```diff
@@ -935,27 +906,37 @@ func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 		sort.Slice(addressesToUpdate, func(i, j int) bool { return bytes.Compare(addressesToUpdate[i][:], addressesToUpdate[j][:]) < 0 })
 		for _, addr := range addressesToUpdate {
 			if obj := s.mutations[addr]; !obj.applied && !obj.isDelete() {
-				s.stateObjects[addr].updateRoot()
+				obj := s.stateObjects[addr] // closure for the task runner below
+				workers.Go(func() error {
```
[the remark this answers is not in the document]
amsanghi: Resolving based on https://github.com/OffchainLabs/go-ethereum/pull/387

eljobe (approved):
LGTM

PR #386, opened 2024-12-17
Revert accidental change to tipReceipient

PlasmaPower:
This was accidentally changed in the geth v1.14.0 merge. It doesn't matter as far as the actual balances go because the tip in Arbitrum is always zero, but it does matter in terms of state accessed by the replay binary and the validator providing that state.


joshuacolvin0 (approved):
LGTM

PR #389, opened 2024-12-20
Include tx compression level in calldata units cache

PlasmaPower:


core/types/transaction.go
```diff
@@ -64,15 +64,54 @@ type Transaction struct {
...
 
 	// caches
 	hash atomic.Pointer[common.Hash]
 	size atomic.Uint64
 	from atomic.Pointer[sigCache]
 }
 
+// GetRawCachedCalldataUnits returns the cached brotli compression level and corresponding calldata units,
+// or (0, 0) if the cache is empty.
+func (tx *Transaction) GetRawCachedCalldataUnits() (uint64, uint64) {
```
tsahee, line 80: can we make this function private (lowercase)?

tsahee (approved):
approved with comment

core/types/transaction.go
PlasmaPower: No, it's used in nitro to retrieve the cached calldata units if any for a metric update

PR #390, opened 2024-12-30
[NIT-3021] paralellise pruning inside accounts

amsanghi:
related to NIT-3021

Pulled in by https://github.com/OffchainLabs/nitro/pull/2898


core/state/pruner/pruner.go
```diff
@@ -427,23 +427,46 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
...
-						processedNodes++
-						if time.Since(threadLastLog) > 5*time.Minute {
-							elapsedTotal := time.Since(startedAt)
-							elapsedThread := time.Since(threadStartedAt)
-							log.Info("traversing trie database - traversing storage trie taking long", "key", key, "elapsedTotal", elapsedTotal, "elapsedThread", elapsedThread, "processedNodes", processedNodes, "threadsRunning", threadsRunning.Load())
-							threadLastLog = time.Now()
-						}
+					}()
+					for isNext {
+						isNext = storageIt.Next(true)
```
tsahee, line 465: I don't think this helps. You still have only one thread traversing all storage, which Is the bottleneck.
Creating parallelism should be done using the parameter in this line:
```
storageTr.NodeIterator(nil)
```
The nil is "from which key to start". Instead of that, have multiple threads start from multiple values and have each stop where the next thread starts. I assume (not sure) that it's enough to give the first part of the key (you obviously don't know the full key at this time)

tsahee (changes requested):

core/state/pruner/pruner.go
amsanghi: Done.

core/state/pruner/pruner.go
```diff
@@ -422,31 +423,57 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
...
+									elapsedThread := time.Since(threadStartedAt)
+									log.Info("traversing trie database - traversing storage trie taking long", "key", key, "elapsedTotal", elapsedTotal, "elapsedThread", elapsedThread, "processedNodes", processedNodes, "threadsRunning", threadsRunning.Load())
+									threadLastLog = time.Now()
+								}
+							}
+							err = startIt.Error()
+							if err != nil {
+								return
+							}
+						}(startStorageIt, nextStorageIt)
```
tsahee, line 475: this will probably behave weird because you use one thread's start as the previous thread's end. 2nd thread advances it's own iterator so previous thread will find it's end iterator is a moving target and both threads will have a lot of shared work

core/state/pruner/pruner.go
```diff
@@ -422,31 +423,57 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
...
-						return
+						go func(startIt, endIt trie.NodeIterator) {
+							threadsRunning.Add(1)
+							defer threadsRunning.Add(-1)
+							var err error
+							defer func() {
+								results <- err
+							}()
+							// Traverse the storage trie, and stop if we reach the end of the trie or the end of the current part
+							for startIt.Next(true) && (endIt == nil || startIt.Hash() == endIt.Hash()) {
```
tsahee, line 454: using an end iterator, even if not shared (see next comment), requires a seek operation to start.
I think it's enough for every thread to have one iterator, the start iterator. End condition should not be on the first byte of the iterator going over the expected value.

core/state/pruner/pruner.go
```diff
@@ -422,31 +423,57 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
...
+										return
+									}
+								}
+								processedNodes++
+								if time.Since(threadLastLog) > 5*time.Minute {
+									elapsedTotal := time.Since(startedAt)
+									elapsedThread := time.Since(threadStartedAt)
+									log.Info("traversing trie database - traversing storage trie taking long", "key", key, "elapsedTotal", elapsedTotal, "elapsedThread", elapsedThread, "processedNodes", processedNodes, "threadsRunning", threadsRunning.Load())
+									threadLastLog = time.Now()
+								}
```
tsahee, line 469: I'd add a Trace print at the end of every thread which prints the first and last value of the iterator, just to see pruning is working as ewxpected.

tsahee (changes requested):

core/state/pruner/pruner.go
amsanghi: Makes sense, changed to just use 1 iterator in a time.

core/state/pruner/pruner.go
amsanghi: Make sense, instead of passing nextiterator, I am not just passing the first hash of the nextiterator, and when we are equal to that, the for loop will stop.

core/state/pruner/pruner.go
```diff
@@ -370,6 +371,11 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
 	for i := 0; i < threads; i++ {
 		results <- nil
 	}
+	// We also create a semaphore for the storage trie traversal, to limit the number of goroutines
+	resultsPerAccount := make(chan error, threads)
```
PlasmaPower, line 375: Why not just use `results` instead of creating a new semaphore? As-is you'll likely end up with 2*threads running, as you have one set of threads for the top level and one set of threads for account storage.

amsanghi: Yeah i am kind of stuck on this and can use your help, so what is happening is earlier there was only one level of multithreading across account, and now its a multilevel threading across accounts and across hashes.

So whats happening is lets say there there are 10 account and 10 threads allowed, 

All the 10 threads are consumed by 10 accounts, and when the time comes to open the a new thread for pruning across hashes there are no more threads left and its a deadlock (earlier this was not an issue since the same thread that was used for the account could be used for pruning the hashes.)

Wanted to get your thoughts on how to handle this case.

PlasmaPower: Why not move the `storageTr.NodeIterator(big.NewInt((i - 1) << 3).Bytes())` call into the inner storage goroutine, and then drop the outer goroutine and only have that happen in the `if !config.ParallelPrune` case? Right now if ParallelPrune is enabled, no significant computation is happening in the accounts goroutine, and this'll collapse it down to always only being one level of goroutines.

amsanghi: Done.

core/state/pruner/pruner.go
```diff
@@ -401,54 +404,121 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
...
+							if err != nil {
+								return
+							}
+							// Traverse the storage trie, and stop if we reach the end of the trie or the end of the current part
+							var startItPath, endItPath []byte
+
+							key := keybytesToHex(big.NewInt((iteration) << 3).Bytes())
+							key = key[:len(key)-1]
+							for startIt.Next(true) {
+								if iteration != 32 && bytes.Compare(startIt.Path(), key) >= 0 {
```
magicxyyz, line 491: shouldn't the second condition be `bytes.Compare(startIt.Path(), key) > 0`?

the NodeIterator starts at the key after the given start key, am I right that with >= we might skip the trie node with `path == key` ?

https://github.com/OffchainLabs/go-ethereum/blob/8070936269ea11801bb91e90dfcf66685cba9fcc/trie/secure_trie.go#L253-L257

core/state/pruner/pruner.go
```diff
@@ -62,6 +63,8 @@ type Config struct {
 	BloomSize      uint64 // The Megabytes of memory allocated to bloom-filter
 	Threads        int    // The maximum number of threads spawned in dumpRawTrieDescendants and removeOtherRoots
 	CleanCacheSize int    // The Megabytes of clean cache size used in dumpRawTrieDescendants
+
+	ParallelPrune bool // Whether to prune in parallel per account
```
magicxyyz, line 67: Can we name it a bit different to specify that it only disables/enables parallelization of account storage traversal? something like `ParralelStorageTraversal` (or something shorter 😄 )

Otherwise, if user sets `ParallelPrune = false` they might expect that there won't be any parallelization of pruning, but traversing accounts trie still will be run in multiple goroutines (if `Threads` > 0, and by default it is `runtime.NumCPU()` if I am not mistaken)

core/state/pruner/pruner.go
```diff
@@ -464,6 +534,17 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
 	return nil
 }
 
+func keybytesToHex(str []byte) []byte {
```
magicxyyz, line 537: instead of copying the keybytesToHex can we export `trie.keyBytesToHex` and use it here? 

e.g. by just adding in `trie/encoding.go`:
```
func KeybytesToHex(str []byte) []byte {
	return keybytesToHex(str)
}
```

core/state/pruner/pruner.go
```diff
@@ -401,54 +404,121 @@ func dumpRawTrieDescendants(db ethdb.Database, root common.Hash, output *stateBl
...
-					defer threadsRunning.Add(-1)
-					var err error
-					defer func() {
-						results <- err
-					}()
-					threadStartedAt := time.Now()
-					threadLastLog := time.Now()
-
-					storageIt, err := storageTr.NodeIterator(nil)
+				if !config.ParallelPrune {
```
magicxyyz, line 407: we are duplicating lots of code here,
can we abstract the storage traversal from start key to end key?

eg. we could have:
```
traverseStorage := func(startKey, endKey []byte) {
 ...
 // if endKey == nil then we skip endKey check
 ...
}
```

and then we could just:
```
if !config.ParallelPrune {
...
    go traverseStorage(nil, nil)
...
} else {
...
    for i := int64(1); i <= 32; i++ {
         ...
         if i == 32 {
             endKey = nil
         }
        go traverseStorage(startKey, endKey)
        ...
    }
...
}

```

magicxyyz (changes requested):

core/state/pruner/pruner.go
amsanghi: Thanks for catching this, updated!

core/state/pruner/pruner.go
amsanghi: Done.

core/state/pruner/pruner.go
amsanghi: Makes sense, though instead of using keys as parameters, I have updated to use number of parts that we are dividing into 1 vs 32 as parameter to remove reductant code.

Let me know if this does not feel right.

magicxyyz (approved):
LGTM

tsahee (approved):
LGTM

PR #393, opened 2025-01-06
fix flatCallTracer onexit out of bound

amsanghi:
Discovered in https://github.com/OffchainLabs/nitro/pull/2676


joshuacolvin0 (approved):
LGTM

PR #396, opened 2025-01-12
Don't modify gas cap if already zero (infinite)

joshuacolvin0:
Also remove out of date comment

Pulled in by https://github.com/OffchainLabs/nitro/pull/2873


internal/ethapi/transaction_args.go
```diff
@@ -490,9 +490,11 @@ func (args *TransactionArgs) ToMessage(baseFee *big.Int, globalGasCap uint64, he
 	}
 	// Arbitrum: raise the gas cap to ignore L1 costs so that it's compute-only
 	if state != nil && !skipL1Charging {
-		// ToMessage recurses once to allow ArbOS to intercept the result for all callers
```
eljobe, line 493: Why did you remove the inline documentation?

eljobe (dismissed):
I'm happy with this change once the documentation comments are restored.

eljobe (approved):
LGTM

PR #398, opened 2025-01-21
Flush triedb function

diegoximenes:
Resolves NIT-3047

- Adds metrics related to triedb size and gcproc
- Adds function to flush triedb to BlockChain

### DIFF
--- a/core/blockchain_arbitrum.go
+++ b/core/blockchain_arbitrum.go
@@ -19,13 +19,31 @@ package core
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/rpc"
 )
+
+func (bc *BlockChain) FlushTrieDB(advanceBlockChainMutex *sync.Mutex, capLimit common.StorageSize) error {
+	advanceBlockChainMutex.Lock()
+	defer advanceBlockChainMutex.Unlock()
+
+	err := bc.triedb.Cap(capLimit)
+	if err != nil {
+		return err
+	}
+
+	bc.gcproc = 0
+	// bc.lastWrite = prevNum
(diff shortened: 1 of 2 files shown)


core/blockchain_arbitrum.go
```diff
@@ -19,14 +19,39 @@ package core
...
+	defer advanceBlockChainMutex.Unlock()
+
+	if bc.triedb.Scheme() == rawdb.HashScheme {
+		err := bc.triedb.Cap(capLimit)
+		if err != nil {
+			return err
+		}
+	}
+
+	err := bc.triedb.Commit(bc.CurrentBlock().Root, true)
```
magicxyyz, line 44: nitpick: I wonder if we couldn't first commit current block root and then cap remaining dirties cache - that way the capLimit parameter will be more precise.
 
 Currently, after `FlushTireDB` triedb size is:
 `sizeAfter = min(sizeBefore, capLimit) - dirtyNodesFromCurrentRootNotFlushedWithCap` 
  
  If we committed first then we'd have:
  `sizeAfter = min(sizeBefore - dirtyNodesFromCurrentRoot, capLimit)`

magicxyyz (approved):
LGTM, added comment to Cap/Commit order

core/blockchain_arbitrum.go
diegoximenes: Done.

core/blockchain_arbitrum.go
```diff
@@ -19,14 +19,41 @@ package core
...
 
+func (bc *BlockChain) FlushTrieDB(advanceBlockChainMutex *sync.Mutex, capLimit common.StorageSize) error {
+	if bc.triedb.Scheme() == rawdb.PathScheme {
+		return nil
+	}
+
+	advanceBlockChainMutex.Lock()
+	defer advanceBlockChainMutex.Unlock()
+
+	err := bc.triedb.Commit(bc.CurrentBlock().Root, true)
```
tsahee, line 41: we don't want to commit currentBlock's root, but the oldest block we have in memory (which is the first entry in triegc)

joshuacolvin0:
Need to make matching PR in nitro repo to pull changes in, and add links between the PRs.

tsahee (approved):
LGTM

PR #399, opened 2025-01-22
Optimize destroying an account with no storage in snapshot diffToDisk

PlasmaPower:
To do this, we get the account info from the disk layer and check if its storage root was empty before trying to iterate over its storage to delete it.

### DIFF
--- a/core/state/snapshot/disklayer.go
+++ b/core/state/snapshot/disklayer.go
@@ -85,10 +85,8 @@ func (dl *diskLayer) account(hash common.Hash, ignoreStale bool) (*types.SlimAcc
 		return nil, nil
 	}
 	account := new(types.SlimAccount)
-	if err := rlp.DecodeBytes(data, account); err != nil {
-		panic(err)
-	}
-	return account, nil
+	err = rlp.DecodeBytes(data, account)
+	return account, err
 }
 
 // AccountRLP directly retrieves the account RLP associated with a particular
--- a/core/state/snapshot/disklayer.go
+++ b/core/state/snapshot/disklayer.go
@@ -76,8 +76,8 @@ func (dl *diskLayer) Stale() bool {
 
 // Account directly retrieves the account associated with a particular hash in
 // the snapshot slim data format.
-func (dl *diskLayer) account(hash common.Hash, ignoreStale bool) (*types.SlimAccount, error) {
-	data, err := dl.accountRLP(hash, ignoreStale)
+func (dl *diskLayer) account(hash common.Hash, evenIfStale bool) (*types.SlimAccount, error) {
+	data, err := dl.accountRLP(hash, evenIfStale)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


core/state/snapshot/snapshot.go
```diff
@@ -560,14 +560,20 @@ func diffToDisk(bottom *diffLayer) *diskLayer {
...
 		// Skip any account not covered yet by the snapshot
-		if base.genMarker != nil && bytes.Compare(hash[:], base.genMarker) > 0 {
+		if errors.Is(err, ErrNotCoveredYet) {
 			continue
 		}
-		// Remove all storage slots
 		rawdb.DeleteAccountSnapshot(batch, hash)
 		base.cache.Set(hash[:], nil)
 
+		if err == nil && (oldAccount == nil || oldAccount.Root == nil) {
```
tsahee, line 571: For the sake of a minimal understandable diff, I'd put here the call to base.account.
Also, either handle err != nil in some way or comment why you don't need to.

core/state/snapshot/disklayer.go
```diff
@@ -76,30 +76,28 @@ func (dl *diskLayer) Stale() bool {
 
 // Account directly retrieves the account associated with a particular hash in
 // the snapshot slim data format.
-func (dl *diskLayer) Account(hash common.Hash) (*types.SlimAccount, error) {
-	data, err := dl.AccountRLP(hash)
+func (dl *diskLayer) account(hash common.Hash, ignoreStale bool) (*types.SlimAccount, error) {
```
tsahee, line 79: thinking alternative names.. not sure they are better, but what do you think about  "evenIfStale" ?

tsahee (dismissed):
code LGTM but made suggestions around documenting/naming

core/state/snapshot/snapshot.go
```diff
@@ -560,14 +560,20 @@ func diffToDisk(bottom *diffLayer) *diskLayer {
 
 	// Destroy all the destructed accounts from the database
 	for hash := range bottom.destructSet {
+		oldAccount, err := base.account(hash, true)
 		// Skip any account not covered yet by the snapshot
-		if base.genMarker != nil && bytes.Compare(hash[:], base.genMarker) > 0 {
+		if errors.Is(err, ErrNotCoveredYet) {
```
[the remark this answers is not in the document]
PlasmaPower: I don't think there's merge conflict concern here because this code seems to be unchanged upstream until it's removed in https://github.com/ethereum/go-ethereum/pull/30752

core/state/snapshot/disklayer.go
PlasmaPower: Sounds good to me, renamed

core/state/snapshot/snapshot.go
PlasmaPower: I've added a warning if err != nil and a comment explaining that it falls through to iterating over and removing storage in case there is any

tsahee (commented):
LGTM

joshuacolvin0 (approved):
LGTM

PR #405, opened 2025-02-01
add is_on_curve_check

0xhashiman:
Check this https://github.com/ethereum/go-ethereum/pull/31100


joshuacolvin0:
This will be included when upstream go-ethereum updates are merged in

PR #406, opened 2025-02-04
Uses finalized block as the freeze threshold

diegoximenes:
Related to NIT-2585

Uses finalized block as the freeze threshold.

Pulled by https://github.com/OffchainLabs/nitro/pull/2927

### DIFF
--- a/core/rawdb/chain_freezer.go
+++ b/core/rawdb/chain_freezer.go
@@ -25,7 +25,6 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/log"
-	"github.com/ethereum/go-ethereum/params"
 )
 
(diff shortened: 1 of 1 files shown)


core/rawdb/chain_freezer.go
```diff
@@ -118,24 +117,23 @@ func (f *chainFreezer) readFinalizedNumber(db ethdb.KeyValueReader) uint64 {
 	return *number
 }
 
-// freezeThreshold returns the threshold for chain freezing. It's determined
-// by formula: max(finality, HEAD-params.FullImmutabilityThreshold).
+// freezeThreshold returns the threshold for chain freezing.
```
tsahee, line 120: It seems like the original function is actually good. It's 100% o.k to have a threshold that won't be frozen even if finalized.
Can we remove the change to geth entirely?

tsahee (changes requested):

PR #412, opened 2025-02-21
Backport changes for Eip 2935

amsanghi:
closes NIT-3105


eljobe (approved):
LGTM

eljobe (changes requested):
Sorry. I know we discussed this, but there isn't a separate Linear ticket for actually modifying the bytecode of the contract of the HistoryStorageAddress to update the size of the ring buffer. So, we either need to do it as part of this PR, or have a separate PR to get that change in as well.

I think, as it is written now, there will only be 8191 historical blocks stored.

amsanghi:
> Sorry. I know we discussed this, but there isn't a separate Linear ticket for actually modifying the bytecode of the contract of the HistoryStorageAddress to update the size of the ring buffer. So, we either need to do it as part of this PR, or have a separate PR to get that change in as well.
> 
> I think, as it is written now, there will only be 8191 historical blocks stored.

Updated the bytecode

eljobe (approved):
LGTM

Issue #418, opened 2025-03-05
Hotfix for Pectra Fork


tebraDev:
# Rationale

Why should this feature exist?
* Shortly after the Pectra fork activation, an issue in deposit contract logs parsing was discovered. Sepolia has a custom version of the deposit contract that also implements ERC-20, and thus unexpected log events may be created by transactions to it. This issue affects all Ethereum execution clients.

Do you have ideas regarding the implementation of this feature?
* `go-ethereum` did already the fix with [this release](https://github.com/ethereum/go-ethereum/releases/tag/v1.15.5)

Are you willing to implement this feature?
* no

tebraDev:
@edfelten can you please have a look at it?

tebraDev:
@eljobe can you check pls - I would like whom to inform to patch the lib

tebraDev:
Any updates?

eljobe:
We'll pick this up when we merge in 1.15.5 from upstream. No need to patch it independently.

PR #420, opened 2025-03-07
Hotfix for Pectra Fork

tebraDev:
> merge of https://github.com/ethereum/go-ethereum/commit/cd78b65cdaa9e00866962e38685c4f93ac9e8444

This resolves a situation on the Sepolia testnet, which has a different deposit contract. The contract on that network emits two kinds of logs, instead of only deposit events like the deposit contract on mainnet. So we need to skip events with mismatched topics.

Part of #418

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -271,12 +271,14 @@ func ProcessParentBlockHash(prevHash common.Hash, vmenv *vm.EVM, statedb *state.
 	statedb.Finalise(true)
 }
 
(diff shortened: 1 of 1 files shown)


tebraDev:
@eljobe @joshuacolvin0 @Tristan-Wilson can some of you check pls?

tebraDev:
@CLAassistant recheck

eljobe:
Hey @tebraDev, thanks for the PR, but we're just planning to catch up with the upstream changes in go-ethereum. We'll pick this up when we pick up the 1.15.5 patch. It doesn't really affect nitro too much as we don't take deposits for consensus clients running on our network.

PR #427, opened 2025-03-16
Introduce Arbitrum_40

eljobe:
This just creates the concept of an arbos 40 release. The variable is actually used in the `nitro` repository, for example, to initialize ArbOwner storage that hold the toggle for the EIP-7623 call data price increase conditionally.

Part of: NIT-3139


joshuacolvin0 (approved):
LGTM

PR #428, opened 2025-03-18
 Fix incorrect type reference and improve error formatting in ABI error generation

Marcofann:
This pull request fixes a logic issue in the `NewError` function where the incorrect `input.Type` reference was used instead of `inputs[i].Type` after potentially modifying the `inputs` slice. This ensures the correct type is always used when generating the string representation of the error.

## Changes
- Corrected type reference from `input.Type` to `inputs[i].Type` in `NewError`.
- Updated the `invalid identifier` error message in `Unpack` to improve readability and consistency.

### DIFF
--- a/accounts/abi/error.go
+++ b/accounts/abi/error.go
@@ -56,9 +56,9 @@ func NewError(name string, inputs Arguments) Error {
 			inputs[i] = input
 		}
 		// string representation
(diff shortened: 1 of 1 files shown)


joshuacolvin0:
These changes will be pulled in when upstream changes are pulled in, not before

PR #429, opened 2025-03-18
Remove the MaxBlobGasPerBlock RPC methods

eljobe:
This solution doesn't work for us, because we cannot be sure that the arbitrum nodes will be talking to L1s that are running our fork of go-ethereum.


joshuacolvin0 (approved):
LGTM

PR #430, opened 2025-03-19
Make it illegal to call Commit with noStorageWiping=true on arbitrum

eljobe:
This is in response to a suggestion on a code review. Essentially, we always want to wipe storage for Arbitrum chains during stateDB commits.


eljobe:
Actually. I don't want to do this. I'm going to drop this commit and fight back in a review comment on https://github.com/OffchainLabs/nitro/pull/3014

PR #431, opened 2025-03-20
Adds a hook to the TxProcessor for EIP-7623

eljobe:
This will allow the Arbitrum implementation of the hook to selectively enable it only when the ArbOwner calls a specific contract.

Part of: NIT-3139


joshuacolvin0 (approved):
LGTM

PR #434, opened 2025-03-25
Add detection of Arbos Version to the evaluation of IsPrauge

eljobe:
This way we can enable the features for Prauge when the Arbos version is
upgraded on chain instead of at some particular block number or block time.

Part of: NIT-3139 and NIT-3173

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -37,6 +37,11 @@ type sigCache struct {
 }
 
 // MakeSigner returns a Signer based on the given chain config and block number.
+//
+// Arbitrum: If the signer is being used in any state-transition affecting code,
+// then this is not the right function to use. Instead, use
+// types.VersionedArbitrumSigner(), which will return the correct signer based
+// on the current block number and time, and active arbitrum version.
 func MakeSigner(config *params.ChainConfig, blockNumber *big.Int, blockTime uint64) Signer {
 	var signer Signer
 	switch {
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -39,17 +39,13 @@ type sigCache struct {
 // MakeSigner returns a Signer based on the given chain config and block number.
 //
 // Arbitrum: If the signer is being used in any state-transition affecting code,
-// then this is not the right function to use. Instead, use
-// types.VersionedArbitrumSigner(), which will return the correct signer based
-// on the current block number and time, and active arbitrum version.
-func MakeSigner(config *params.ChainConfig, blockNumber *big.Int, blockTime uint64) Signer {
+// then the active arbos version must be passed in. In places which don't affect
+// the state transition function, it is okay to pass in
+// params.MaxArbosVersionSupported.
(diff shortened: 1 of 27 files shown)


rauljordan (approved):

rauljordan (dismissed):

rauljordan (dismissed):

core/types/transaction_signing.go
```diff
@@ -37,10 +37,19 @@ type sigCache struct {
 }
 
 // MakeSigner returns a Signer based on the given chain config and block number.
+//
+// Arbitrum: If the signer is being used in any state-transition affecting code,
+// then this is not the right function to use. Instead, use
+// types.VersionedArbitrumSigner(), which will return the correct signer based
+// on the current block number and time, and active arbitrum version.
 func MakeSigner(config *params.ChainConfig, blockNumber *big.Int, blockTime uint64) Signer {
```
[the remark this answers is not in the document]
eljobe: Sold! And Done.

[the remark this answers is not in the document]
eljobe: Done.

eljobe:
Assigning to @tsahee to provide `design-approved` label if he's satisfied with the latest changes.

internal/ethapi/simulate.go
```diff
@@ -168,9 +168,10 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 			}
 		}
 	}
-	if sim.chainConfig.IsCancun(header.Number, header.Time, types.DeserializeHeaderExtraInformation(header).ArbOSFormatVersion) {
+	arbOSVersion := types.DeserializeHeaderExtraInformation(header).ArbOSFormatVersion
```
tsahee, line 171: header here is an output, not input. So you should use arbos version in parent and call it "parentArbosVersion" or something similar.
This will mean simulator will fail if it's asked to simulate a block with an arbos-upgrade and a Prague-only tx in the same block.. we can mention it in a comment or something but I don't think we care.

tsahee (changes requested):

internal/ethapi/simulate.go
eljobe: Okay. This is done.

But, also! Since we don't support EIP-7685 general purpose Consensus layer requests for arbitrum, those code blocks are also explicitly disabled now in this PR.

This was causing problems in some of our tests in the `nitro` repo because there is some subtlety around getting the `RequestsHash` field populated correctly when there are no corresponding requests to handle in a post-Prague block.

tsahee: unused arbosVersion var

tsahee (approved):
two minor comments, can be done in separate PR

internal/ethapi/simulate.go
eljobe: It's not unused. It's used on line 173. This was the behavior of the "IsCancun" check before I started working on this PR. The outer check looked at the version from the header, and the inner check used the version from the parent.

Tristan-Wilson (approved):

PR #442, opened 2025-04-07
add random offset to trie-time-limit

magicxyyz:
Part of NIT-3221

This PR adds option to random offset each period of state trie commits in a full node. This is specifically useful when running a fleet of full nodes and wanting to better distribute commit operations to avoid all nodes committing state trie at the same time.
`TrieTimeLimitRandomOffset` config is added to configure the range of the randomized offset.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"math"
 	"math/big"
-	"math/rand/v2"
 	"runtime"
 	"slices"
 	"strings"
@@ -163,7 +162,7 @@ type CacheConfig struct {
 	// Arbitrum: configure GC window
 	TriesInMemory             uint64        // Height difference before which a trie may not be garbage-collected
 	TrieRetention             time.Duration // Time limit before which a trie may not be garbage-collected
-	TrieTimeLimitRandomOffset time.Duration // Range of random offset of the commit due to TrieTimeLimit period (randomizes initial gcproc value)
+	TrieTimeLimitRandomOffset time.Duration // Range of random offset of each commit due to TrieTimeLimit period
 
 	MaxNumberOfBlocksToSkipStateSaving uint32
 	MaxAmountOfGasToSkipStateSaving    uint64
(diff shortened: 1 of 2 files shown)


core/blockchain.go
```diff
@@ -376,6 +379,10 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 	bc.prefetcher = newStatePrefetcher(chainConfig, bc.hc)
 	bc.processor = NewStateProcessor(chainConfig, bc.hc)
 
+	if cacheConfig.TrieTimeLimitRandomOffset > 0 {
+		bc.gcproc = time.Duration(rand.Int64N(int64(cacheConfig.TrieTimeLimitRandomOffset)))
```
tsahee, line 383: need to also set whenever setting gcproc to 0

tsahee (changes requested):

core/blockchain.go
tsahee: specifically:
* FlushTrieDB
* writeBlockchainWithState

magicxyyz: I don't think we need to randomly offset the gcproc after each commit. Spreading the commits once should make it more predictable (if it spreads nicely at start it shouldn't drift towards time collision soon) and "TrieTimeLimit" will be still working the same as before.

Or do you think that it be better to random offset each time? so if we are unlucky with the first distribution, each next one will be again randomized? 
If you think it's better to randomize each time, I'll implement it a bit differently, so as to keep gcproc intact so the logged value of processing time since last commit will be accurate.

tsahee: I think we shoud randomize each time, to make sure we never enter a steady-state of many synchronizations.
One important point is that we should randomize only each time that we flush (randomize only on startup has some potential for weird steady state, randomize on every time we measure gcpro - wil always flush when random selected will be close to zero).
I accept your point about wanting to keep gcproc's meaning.
Maybe best would be to add a separate "randomOffset" field that is done separately.

magicxyyz: done, added `bc.gcprocRandOffset` which is added to `gcproc` during check against `flushInterval` so the full trie flush still happens no later then `trie-time-limit`, but can happen up to `trie-time-limit-random-offset` earlier. The reasoning behind it is that `trie-time-limit` will still be the maximum time we want to spend re-executing in case of unclean shutdown.

tsahee (approved):
LGTM

PR #447, opened 2025-04-28
ArbOS 41: Changes related to Mint and Burn native token

diegoximenes:
Related to NIT-3298

Changes related to Mint and Burn native token.

It adds:
- BalanceIncreaseMintNativeToken and BalanceDecreaseBurnNativeToken BalanceChangeReason
- ExpectBalanceMint to StateDB

Pulled by https://github.com/OffchainLabs/nitro/pull/3186


eljobe (approved):
This looks good to me so far.

eljobe:
Oh. Can you add more of a description and context to the PR description so we can remember why we made this change later?

params/config_arbitrum.go
```diff
@@ -37,13 +37,14 @@ const ArbosVersion_30 = uint64(30)
...
 const ArbosVersion_40 = uint64(40)
+const ArbosVersion_41 = uint64(41)
 
 const ArbosVersion_FixRedeemGas = ArbosVersion_11
 const ArbosVersion_Stylus = ArbosVersion_30
 const ArbosVersion_StylusFixes = ArbosVersion_31
 const ArbosVersion_StylusChargingFixes = ArbosVersion_32
-const MaxArbosVersionSupported = ArbosVersion_40
-const MaxDebugArbosVersionSupported = ArbosVersion_40
+const MaxArbosVersionSupported = ArbosVersion_41
```
tsahee, line 46: I think we don't want to increase MaxArbosVersionSupported, only increase the Debug version.
I think we're fine with other people doing that change themselbves if they want it
@eljobe

core/tracing/gen_balance_change_reason_stringer.go
```diff
@@ -24,15 +24,44 @@ func _() {
...
+	_BalanceChangeReason_index_0 = [...]uint16{0, 11, 41, 71, 96, 125, 160, 181, 205, 231, 256, 264, 276, 303, 330, 361, 367}
+	_BalanceChangeReason_index_1 = [...]uint16{0, 18, 40, 67, 93, 116, 141, 160, 181, 203, 228, 242, 267, 292, 321, 342, 369, 399, 429}
+)
 
 func (i BalanceChangeReason) String() string {
-	if i >= BalanceChangeReason(len(_BalanceChangeReason_index)-1) {
+	switch {
+	case i <= 15:
+		return _BalanceChangeReason_name_0[_BalanceChangeReason_index_0[i]:_BalanceChangeReason_index_0[i+1]]
+	case 128 <= i && i <= 145:
```
tsahee, line 61: nice!
but can we make it less magic-numbery?

tsahee (changes requested):

core/tracing/gen_balance_change_reason_stringer.go
eljobe: This is a generated file. I don't think we need to worry about the magic-numberiness of it. Or am I missing something?

eljobe (dismissed):

tsahee (approved):
LGTM

tsahee:
Converted it to draft to prevent mistaken merge before everything else is settled. Feel free to publish if nitro side ready to be merged

params/config.go
```diff
@@ -344,6 +344,12 @@ var NetworkNames = map[string]string{
 	HoleskyChainConfig.ChainID.String(): "holesky",
 }
 
+// Arbitrum
+// ArbOSInit defines some initialization values for ArbOS state.
+type ArbOSInit struct {
+	NativeTokenSupplyManagementEnabled bool `json:"nativeTokenSupplyManagementEnabled"`
```
eljobe, line 350: I think this still needs to be a timestamp. Because, for chains where the feature is currently disabled, we want there to be a 7-day delay before it can be enabled. So, we need to implement it as a time after which the mint and burn precompiles can be called.

eljobe (changes requested):

params/config.go
eljobe: Done.

eljobe (dismissed):
LGTM

eljobe (approved):

PR #448, opened 2025-04-30
Avoid overflow when adding additional gas

joshuacolvin0:
Pulled in by https://github.com/OffchainLabs/nitro/pull/3192

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -721,7 +721,9 @@ func applyMessage(ctx context.Context, b Backend, args TransactionArgs, state *s
 		if err != nil {
 			return nil, err
 		}
-		gp.AddGas(postingGas)
+		if gp.Gas() < gomath.MaxUint64-postingGas {
+			gp.AddGas(postingGas)
+		}
 	}
 	if msg.GasLimit > gp.Gas() {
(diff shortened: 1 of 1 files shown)


internal/ethapi/api.go
```diff
@@ -721,7 +721,9 @@ func applyMessage(ctx context.Context, b Backend, args TransactionArgs, state *s
 		if err != nil {
 			return nil, err
 		}
-		gp.AddGas(postingGas)
+		if gp.Gas() < gomath.MaxUint64-postingGas {
```
amsanghi, line 724: Shouldn't we add an else as well something like 
```
if gp.Gas() < gomath.MaxUint64-postingGas {
  gp.AddGas(postingGas)
} else {
  gp.SetGas(gomath.MaxUint64)
}
```

Also should we consider moving this logic inside of `AddGas` ?

amsanghi (approved):

internal/ethapi/api.go
joshuacolvin0: Idea is to minimize the locations that we patch geth, so don't want to move logic to `AddGas`. 
No need to use the `else` because `postingGas` isn't going to be large, so so that means `gp.Gas()` is already huge.

PR #451, opened 2025-05-06
Allow nitro nodes to forward requests based on block number to archive nodes

ganeshvanahalli:
This PR allows forwarding of requests to archive nodes based on block numbers thus allowing splitting of archive databases for better manageability. Configured via `archive-redirects` option which is a list of `ArchiveRedirectConfig` consisting of fields- `url`, `timeout` and `last-block` (block up to which this archive node can service requests). 

User can also configure via command line using `archive-redirects-list` which is a string that should be in json format consisting array of above archive node configs.
```
--execution.rpc.archive-redirects-list=`[{"url": "<>", "timeout":<in nano seconds>, "last-block": 100}, {"url":"<>}]`
```

part of NIT-3211

### DIFF
--- a/arbitrum/archivefallbackclients.go
+++ b/arbitrum/archivefallbackclients.go
@@ -1,0 +1,71 @@
+package arbitrum
+
+import (
+	"math/rand"
+	"sort"
+
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type lastBlockAndClient struct {
+	lastBlock uint64
+	client    types.FallbackClient
+}
+
+type archiveFallbackClientsManager struct {
+	lastBlockAndClients []*lastBlockAndClient
+}
+
+func newArchiveFallbackClientsManager(archiveRedirects []ArchiveRedirectConfig) (*archiveFallbackClientsManager, error) {
+	manager := &archiveFallbackClientsManager{}
+	for _, archiveConfig := range archiveRedirects {
+		fallbackClient, err := CreateFallbackClient(archiveConfig.URL, archiveConfig.Timeout, true)
+		if err != nil {
+			return nil, err
+		}
+		if fallbackClient == nil {
+			continue
+		}
+		manager.lastBlockAndClients = append(manager.lastBlockAndClients, &lastBlockAndClient{
+			lastBlock: archiveConfig.LastBlock,
+			client:    fallbackClient,
+		})
+	}
+	if len(manager.lastBlockAndClients) == 0 {
+		return nil, nil
+	}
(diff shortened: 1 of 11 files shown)


joshuacolvin0:
Requests can come in to any of the nodes, so each node will need to be able to determine which node to forward requests to if needed.  So a node that has information for blocks 10,000-20,000 might forward request for block 500 to a node1, and block 21,000 to node 3.

arbitrum/config.go
```diff
@@ -39,6 +41,27 @@ type Config struct {
...
+
+	ArchiveRedirects     []ArchiveRedirectConfig `koanf:"archive-redirects"`
+	ArchiveRedirectsList string                  `koanf:"archive-redirects-list"`
+}
+
+type ArchiveRedirectConfig struct {
+	URL       string        `koanf:"url"`
+	Timeout   time.Duration `koanf:"timeout"`
+	LastBlock uint64        `koanf:"last-block"`
+}
```
Tristan-Wilson, line 53: Do we see ourselves having archive nodes with an "EarliestBlock" too? If that's the case then we should have that info too. We'd also have to change the pool function.

Tristan-Wilson (changes requested):

arbitrum/config.go
ganeshvanahalli: Nope we wont be supporting `EarliestBlock` instead it is assumed the blocks presence is contiguous i.e if i give configs with last blocks as nodeA=100 and nodeB=200 then requests for blocks from classic.last-block to 100 will go to nodeA and 101 to 200 will go to nodeB

arbitrum/archivefallbackclients.go
```diff
@@ -0,0 +1,53 @@
...
+func (a *archiveFallbackClientsManager) lastAvailableBlock() uint64 {
+	return a.lastBlockAndClients[len(a.lastBlockAndClients)-1].lastBlock
+}
+
+func (a *archiveFallbackClientsManager) fallbackClient(blockNum uint64) types.FallbackClient {
+	for _, lastBlockAndClient := range a.lastBlockAndClients {
+		if blockNum <= lastBlockAndClient.lastBlock {
+			return lastBlockAndClient.client
+		}
+	}
```
Tristan-Wilson, line 51: Could we have it randomly select within clients that have the same last block? Otherwise it'll always send to the same one.

ganeshvanahalli: addressed it

Tristan-Wilson (approved):

eljobe (approved):

PR #456, opened 2025-05-09
Revert setFeeDefaults to original code from geth

amsanghi:
Pulled in https://github.com/OffchainLabs/nitro/pull/3228


tsahee (approved):
LGTM

PR #461, opened 2025-05-20
Adds ProcTimeBeforeFlush function to BlockChain

diegoximenes:
Related to NIT-3231 and NIT-3232

Adds ProcTimeBeforeFlush function to BlockChain.

Pulled by https://github.com/OffchainLabs/nitro/pull/3253


core/blockchain.go
```diff
@@ -1555,6 +1555,16 @@ func (bc *BlockChain) writeKnownBlock(block *types.Block) error {
 	return nil
 }
 
+func (bc *BlockChain) TimeBeforeFlush() (time.Duration, error) {
```
magicxyyz, line 1558: nitpick: I think we should rename the method to something like `ProcTimeBeforeFlush` as the returned duration is not absolute time but the processing time accumulated in gcproc. Gcproc is a hugely different measure e.g. 1 h of proc time can take ~12 hours of real time.

magicxyyz (changes requested):
Looks good, added one nitpick comment for the method name

magicxyyz (approved):
LGTM

eljobe:
Just assigning this to @diegoximenes until the corresponding nitro PR has had the review comments addressed.

PR #463, opened 2025-05-23
Disable the i386 lint action

eljobe:
This was added at some point, but we don't have self-hosted CI runners in the OffchainLabs pool of runners, so, these actions never get picked up.

Linting is already covered in ci.yml, anyway.


joshuacolvin0 (approved):
LGTM

PR #465, opened 2025-05-30
change synchronisation in hashdb commit to allow concurrent reads

magicxyyz:
This PR modifies hashdb.commit operation:
* nodes removal from dirties is delayed until all writes to disk are performed
* read lock is used for the fist part of commit operation
* hashdb write lock is acquired only when removing nodes from dirties

part of NIT-3414


magicxyyz:
closed in favour of a less invasive solution that doesn't modify hashdb locking

PR #468, opened 2025-06-13
return missing targets when handling wasm activations in statedb

magicxyyz:
part of NIT-3358

In this PR:
- StateDB method that gets asmMap (target to assembly map), for given module hash and target list,  is modified to return list of targets for which assembly data was not found
- In `ActivateWasm`, safety check is added that ensures that newly activated asmMaps contain entries for the same target list
- Cleans up error and "not found" handling when reading from wasm store


tsahee (commented):
small comment

core/state/statedb_arbitrum.go
```diff
@@ -272,9 +311,9 @@ func (s *StateDB) RecordProgram(targets []rawdb.WasmTarget, moduleHash common.Ha
 		// nothing to record
 		return
 	}
-	asmMap, err := s.TryGetActivatedAsmMap(targets, moduleHash)
-	if err != nil {
-		log.Crit("can't find activated wasm while recording", "modulehash", moduleHash, "err", err)
+	asmMap, missingTargets, err := s.ActivatedAsmMap(targets, moduleHash)
+	if err != nil || len(missingTargets) > 0 {
+		log.Crit("can't find activated wasm while recording", "modulehash", moduleHash, "err", err, "missingTargets", missingTargets)
```
gligneul, line 316: log.Crit means we exit the node, correct? If so, it would be nice to give a hint to solve the issue.

gligneul (dismissed):
Looks good, just minor comments.

core/state/statedb_arbitrum.go
```diff
@@ -74,48 +74,87 @@ func NewStylusPrefix(dictionary byte) []byte {
...
+				}
+			}
+		}
+		if inconsistent {
+			var previousTargets, newTargets []rawdb.WasmTarget
+			for target := range previouslyActivated {
+				previousTargets = append(previousTargets, target)
+			}
+			for target := range asmMap {
+				previousTargets = append(previousTargets, target)
```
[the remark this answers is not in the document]
magicxyyz: thanks, used the `slices.Collect(maps.Keys())`, those are nice!

core/state/statedb_arbitrum.go
magicxyyz: yes, log.Crit calls os.Exit and it may corrupt databases

after log.Crit there's no good way to recover, except from restoring node from a snapshot, and that applies to all log.Crits

this check should fail only in case of a bug, so maybe it's not that bad to kill the node here

alternatively we could log.Error here and the issue of missing targets would propagate to validation workers that would fail because of the missing targets

core/state/journal_arbitrum.go
```diff
@@ -59,8 +59,8 @@ type EvictWasm struct {
 }
 
 func (ch EvictWasm) revert(s *StateDB) {
-	asm, err := s.TryGetActivatedAsm(rawdb.LocalTarget(), ch.ModuleHash) // only happens in native mode
-	if err == nil && len(asm) != 0 {
+	asm := s.ActivatedAsm(rawdb.LocalTarget(), ch.ModuleHash) // only happens in native mode
```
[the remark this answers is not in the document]
magicxyyz: tried to reword it, please let me know if I should change it further

core/state/statedb_arbitrum.go
tsahee: try to nevere use log.Crit.
Using panic is much better, because it'll only kill an rpc if we're inside such context.
That's only if you cannot return an error, of course.

tsahee: from an initial look - it seems like we can change the function to return an error here

tsahee (changes requested):
just that log.Crit

tsahee (approved):

PR #470, opened 2025-06-17
Create codeql.yml

eljobe:
This is needed because there isn't actually java-kotlin code in the repo that we want CodeQL to be running for. The default configuration includes all of the languages, so this fails the CodeQL run.


eljobe:
@joshuacolvin0, looking for `ddesign-approved`

PR #473, opened 2025-06-20
Return 0 in CalcBlobFee for arbitrum chains since arbitrum does not have blobs

amsanghi:
Related to NIT-3467
Pulled in https://github.com/OffchainLabs/nitro/pull/3343


joshuacolvin0 (approved):
LGTM

PR #478, opened 2025-06-26
Add hash scheme as default for arbitrum eth config

amsanghi:
If node starts with empty db (eg syncing sepolia from genesis) will start by default with Pathdb chosen, so to avoid that adding hash scheme as default here.


eljobe (changes requested):
Why?
Can you either add the answer in the PR description, or inline in a comment?

eljobe (approved):

PR #479, opened 2025-06-26
Record Block information as Part of Gas Dimension Live Tracer

relyt29:
Adds some additional info (block base fee) for the live tracer to trace, and saves the block information to protobuf format rather than raw txt file

Note: the target of this PR is PR #476, rather than master


eljobe (dismissed):

eljobe:
Hey @relyt29, sorry. I went ahead and merged #476 before realizing that you wanted to get this one in on top of it.
So, we'll need a separate PR to pull this into the `nitro` repo.

PR #482, opened 2025-07-02
Add a type for modeling a Resource Constraint

gligneul:
This PR introduces the package multigas that contains the code related to multi-dimensional gas. It adds the ResourceKind enum that represents a gas resource, and the ResourceConstraints struct that represents a set of constraints.

Close NIT-3487

### DIFF
--- a/arbitrum/multigas/constraint_test.go
+++ b/arbitrum/multigas/constraint_test.go
@@ -1,0 +1,66 @@
+package multigas
+
+import (
+	"testing"
+	"time"
+)
+
+func TestResourceConstraints(t *testing.T) {
+	rc := NewResourceConstraints()
+
+	const (
+		minuteSecs = 60
+		daySecs    = 24 * 60 * 60
+		weekSecs   = 7 * daySecs
+		monthSecs  = 30 * daySecs
+	)
+
+	// Adds a few constraints
+	rc.SetConstraint(ResourceKindComputation, minuteSecs, 5_000_000*minuteSecs)
+	rc.SetConstraint(ResourceKindComputation, weekSecs, 3_000_000*weekSecs)
+	rc.SetConstraint(ResourceKindHistoryGrowth, monthSecs, 1_000_000*monthSecs)
+	if len(rc[ResourceKindComputation]) != 2 {
+		t.Fatalf("expected 2 computation constraints")
+	}
+	if rc[ResourceKindComputation][minuteSecs].period != time.Duration(minuteSecs)*time.Second {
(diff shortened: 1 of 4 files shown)


arbitrum/multigas/resources.go
```diff
@@ -0,0 +1,15 @@
+package multigas
+
+// ResourceKind represents a dimension for the multi-dimensional gas.
+type ResourceKind uint8
+
+const (
+	ResourceKindComputation ResourceKind = iota
```
eljobe, line 7: Can we make it a habit for our enums to use the first value, zeroth value as "Unknwon"
Or, in this example ResourceKindUnknown

Also, maybe if we put this type in a "kind" subpackage, we could just call it Resource and have names like kind.Unknown, kind.StorageAccess? Just an idea. Take it or leave it.

eljobe (changes requested):

arbitrum/multigas/resources.go
gligneul: I agree with the first bit. I don't like creating a package named kind because it is too generic.

arbitrum/multigas/constraint_test.go
```diff
@@ -0,0 +1,53 @@
+package multigas
+
+import (
+	"testing"
+	"time"
+)
+
+func assertEqual[T comparable](t *testing.T, description string, got T, want T) {
+	t.Helper()
```
eljobe, line 9: Optional:

This is definitely a matter of preference, but I don't like helper functions like this one. I think it makes it harder to follow the flow of the places where the tests could have failed.

Also, I don't love that it always uses `t.Fatalf` when I believe there are places in the test where `t.Errorf` would make more sense. For example. Sure, fail the test if the wrong number of constraints were created because later tests aren't going to be able to run. But, when checking of a particular constraint has the right value, we should probably get errors for each different constraint that has the wrong value, and `t.Errorf` would give us that.

arbitrum/multigas/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package multigas defines multi-dimensional gas for the EVM.
```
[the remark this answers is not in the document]
gligneul: I like it because it's easier to find where the documentation is. I can remove it if you want.

eljobe (approved):

arbitrum/multigas/doc.go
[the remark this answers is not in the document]
gligneul: Created nitro PR https://github.com/OffchainLabs/nitro/pull/3374

PR #483, opened 2025-07-04
Add more methods to MultiGas struct

gligneul:


gligneul:
> @gligneul, are you ready for review? If so, can you click, "Ready for review", give it a proper PR description and assign this to me?
> If not, what's still outstanding?

There are quite a few things missing. I will open it soon.

MishkaRogachev (dismissed):
Overall everything looks good, there is one small suggestion:
```
func (mg *MultiGas) ToLegacyGas() uint64 {
   var sum uint64
   for _, value := range mg.gas {
   	sum = math.SafeAdd(sum, value)
   }
   return sum
}
```
it will help me in the tests

eljobe:
> Overall everything looks good, there is one small suggestion:
> 
> ```
> func (mg *MultiGas) ToLegacyGas() uint64 {
>    var sum uint64
>    for _, value := range mg.gas {
>    	sum = math.SafeAdd(sum, value)
>    }
>    return sum
> }
> ```
> 
> it will help me in the tests

I would like to not call such a method `ToLegacyGas` because it makes it seem like we are no longer using that kind of gas. But, in fact, single-dimensional gas is still a valid concept in the system.
Let's go with `SingleDimGas()` or `SingleGas()` or `Aggregate()` or something like that.

eljobe (approved):

eljobe (approved):

MishkaRogachev (approved):

PR #496, opened 2025-07-23
Update gasFunc signature to return only multi-dimensional gas

MishkaRogachev:
For NIT-3543

- [x] Update the gasFunc API to return only `(*multigas.MultiGas, error)`
- [x] For uninstrumented methods, temporarily return `ResourceKindUnknown` (added TODO comments)

### DIFF
--- a/arbitrum/multigas/resources.go
+++ b/arbitrum/multigas/resources.go
@@ -36,6 +36,11 @@ func NewMultiGas(kind ResourceKind, amount uint64) *MultiGas {
 	return mg
 }
 
+// UnknownGas returns a MultiGas initialized with unknown gas.
+func UnknownGas(amount uint64) *MultiGas {
+	return NewMultiGas(ResourceKindUnknown, amount)
+}
+
 // ComputationGas returns a MultiGas initialized with computation gas.
 func ComputationGas(amount uint64) *MultiGas {
 	return NewMultiGas(ResourceKindComputation, amount)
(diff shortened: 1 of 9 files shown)


arbitrum/multigas/resources.go
```diff
@@ -36,6 +36,11 @@ func NewMultiGas(kind ResourceKind, amount uint64) *MultiGas {
 	return mg
 }
 
+// UnknownGas returns a MultiGas initialized with unknown gas.
+func UnknownGas(amount uint64) *MultiGas {
```
eljobe, line 40: I have an idea.
For code safety, since this is meant to be a temporary thing:
What if we had the library panic if there is ever any unknown gas AND some other kind of gas on a MultiGas at the same time?

Essentially, this would represent a coding error and should not be allowed. Either the MultiGas has some "Unknown" single-dimensional gas that it's just currying through since we haven't instrumented the gasFunc yet, or it's going to have 0 "Unknown" gas and some amount of other types of gas.

Maybe we should enforce that through a panic or at least returning an error. I kind of think that panic might be the right choice since we don't want the STF to continue if this error occurs.

eljobe (changes requested):

arbitrum/multigas/resources.go
gligneul: That shouldn't work out because there are parts of the code that already use multigas and others that don't. For instance, memoryGasCost already returns the correct amount of multigas, but we need to add up in places where it wasn't instrumented yet. Instead of making this change now, we could create a ticket to remove this UnknownGas function later, ensuring it is not called anywhere.

PR #497, opened 2025-07-25
Instrument multi-gas for constant-gas costs

gligneul:
I went through the opcodes in the tracer, and the only one that contains a constant gas cost that is not computation is self-destruct (4900 StateAccess and 100 computation).

Close NIT-3486

### DIFF
--- a/core/vm/gas.go
+++ b/core/vm/gas.go
@@ -57,6 +57,12 @@ func callGas(isEip150 bool, availableGas, base uint64, callCost *uint256.Int) (u
 
 // constantMultiGas returns the constant multi-gas cost of an opcode.
 func constantMultiGas(cost uint64, op OpCode) *multigas.MultiGas {
+	// SELFDESTRUCT is a special case because it charges for storage access but it isn't
+	// dependent on any input data. We charge a small computational cost for warm access like
+	// other multi-dimensional gas opcodes, and the rest is storage access to delete the
+	// contract from the database.
+	// Note we only need to cover EIP150 because it the current cost, and SELFDESTRUCT cost was
+	// zero previously.
 	if op == SELFDESTRUCT && cost == params.SelfdestructGasEIP150 {
 		return multigas.MultiGasFromMap(map[multigas.ResourceKind]uint64{
 			multigas.ResourceKindComputation:   params.WarmStorageReadCostEIP2929,


arbitrum/multigas/resources_test.go
```diff
@@ -6,22 +6,24 @@ import (
 )
 
 func TestMultiGas(t *testing.T) {
-	t.Run("Test constructor", func(t *testing.T) {
-		// Test ZeroGas
+	t.Run("TestZeroGas", func(t *testing.T) {
```
eljobe, line 9: I've seen this pattern a few times in test code, and I really don't understand it. (Not saying it's wrong. I'm genuinely trying to understand if there is an advantage to structuring the test code this way.)

But, why do you prefer having all of these `t.Run("TestNameThatLooksLikeAFunction", func())` calls inside a parent test function, rather than just having separate test functions for each case?

```go
func TestZeroGas(t *testing.T) {
}

func TestComputationGas(t *testing.t) {
}
...
```

core/vm/gas.go
```diff
@@ -52,3 +54,14 @@ func callGas(isEip150 bool, availableGas, base uint64, callCost *uint256.Int) (u
 
 	return callCost.Uint64(), nil
 }
+
+// constantMultiGas returns the constant multi-gas cost of an opcode.
+func constantMultiGas(cost uint64, op OpCode) *multigas.MultiGas {
+	if op == SELFDESTRUCT && cost == params.SelfdestructGasEIP150 {
```
eljobe, line 60: Can we have some commentary about why SELFDESTRUCT is special?
Maybe it would be better-suited to PR description, or updating the ADR-002 with information that will capture the answer to the question "why?" for posterity?

arbitrum/multigas/resources_test.go
eljobe: This one is really just a discussion point, and not a change I need.

gligneul: I don't have a strong reason. In this case, it evolved to be that way, but it could easily be separate functions.

core/vm/gas.go
gligneul: I was following the tracer here, but in retrospect, it is not clear why the base cost of self-destruct is multi-dimensional instead of just being computation.

eljobe: Sorry. I'm not saying that it should't be multi-dimensional. I'm guessing that Tyler and Ben thought this through, and that's why the tracer sets it up that way. But, can we write down either directly as inline documentation comments or as part of ARD-002.md the rationale so we don't later forget why it's special cased?

gligneul: After reviewing the opcode it makes sense to be multi-dimensional. I will add a comment.

PR #500, opened 2025-07-29
FIx computeHoldGas handling

joshuacolvin0:
related to NIT-3634
pulled in by https://github.com/OffchainLabs/nitro/pull/3431

refunding gas held to limit eth_call execution was happening after peakGas calculation which can cause unexpected values


rauljordan (approved):
I don't see why this would be a problem so far because HeldGas is only non-zero for eth_call. Also, there is only one invocation of calcRefund which is the few lines below. If there were more invocations of it elsewhere, I would be more wary. The only difference here is in in the value of peekGasUsed which has a floor value anyways (enforced a few lines down)

PR #507, opened 2025-08-14
Instrument the SLOADopcode

alexandrosfilios:


core/state/access_events.go
```diff
@@ -151,6 +153,36 @@ func (ae *AccessEvents) AddTxDestination(addr common.Address, sendsValue bool) {
 	ae.touchAddressAndChargeGas(addr, zeroTreeIndex, utils.CodeHashLeafKey, false)
 }
 
+// SlotMultigas returns the amount of gas to be charged for a cold storage access.
+func (ae *AccessEvents) SlotMultigas(addr common.Address, slot common.Hash, isWrite bool) (*multigas.MultiGas, error) {
```
MishkaRogachev, line 157: I would suggest fixing the existing functions in `access_events.go` instead of adding a new so that it will be easier to track changes in the next update from the go-ethereum upstream

core/vm/operations_verkle.go
```diff
@@ -25,20 +25,25 @@ import (
 )
 
 func gasSStore4762(evm *EVM, contract *Contract, stack *Stack, mem *Memory, memorySize uint64) (*multigas.MultiGas, error) {
-	gas := evm.AccessEvents.SlotGas(contract.Address(), stack.peek().Bytes32(), true)
-	if gas == 0 {
-		gas = params.WarmStorageReadCostEIP2929
```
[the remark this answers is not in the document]
MishkaRogachev: Yes, I inadvertently overlooked that

core/state/access_events.go
```diff
@@ -152,33 +154,51 @@ func (ae *AccessEvents) AddTxDestination(addr common.Address, sendsValue bool) {
...
 	}
-	return gas
+	return gas, nil
+}
+
+// touchAddressAndChargeGas adds any missing access event to the access event list, and returns the cold
+// access cost to be charged, if need be.
+func (ae *AccessEvents) touchAddressAndChargeGas(addr common.Address, treeIndex uint256.Int, subIndex byte, isWrite bool) uint64 {
+	multiGas, err := ae.touchAddressAndChargeMultigas(addr, treeIndex, subIndex, isWrite)
+	if err != nil {
```
MishkaRogachev, line 198: nit: log an error just in case

MishkaRogachev (approved):
Nice!

core/state/access_events.go
```diff
@@ -152,33 +154,51 @@ func (ae *AccessEvents) AddTxDestination(addr common.Address, sendsValue bool) {
...
 	stemRead, selectorRead, stemWrite, selectorWrite, selectorFill := ae.touchAddress(addr, treeIndex, subIndex, isWrite)
 
-	var gas uint64
-	if stemRead {
-		gas += params.WitnessBranchReadCost
+	gas := multigas.ZeroGas()
+	// Reading includes witness calculation (computation), and storage access
+	// The computation cost for the witness calculation is negligible compared to the storage access cost
+	if stemRead && gas.SafeIncrement(multigas.ResourceKindStorageAccess, params.WitnessBranchReadCost) {
+		return nil, fmt.Errorf("failed to increment gas for branch read due to overflow")
```
gligneul, line 171: I would not check for overflow in this function because it is impossible to happen. Also, I would prefer to keep the code structure similar to what it was before to make future merges easier.

gligneul (changes requested):

gligneul (approved):

gligneul (approved):
LGTM

eljobe (approved):

PR #523, opened 2025-08-28
Revert EIP-7825 (moved into nitro)

ganeshvanahalli:
This PR removes EIP-7825 implementation added via upstream merge of v1.16.2 as this EIP is implemented on the nitro side for simplicity

Part of NIT-3644

### DIFF
--- a/tests/state_test.go
+++ b/tests/state_test.go
@@ -57,9 +57,6 @@ func initMatcher(st *testMatcher) {
 	// Broken tests:
 	// EOF is not part of cancun
 	st.skipLoad(`^stEOF/`)
-
-	// Skip all EIP-7825 gas limit cap tests as EIP-7825 is not implemented in nitro's geth fork
-	st.skipLoad(`eip7825_transaction_gas_limit_cap`)
 }
 
 func TestState(t *testing.T) {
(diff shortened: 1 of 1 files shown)


cmd/evm/internal/t8ntool/transaction.go
```diff
@@ -183,9 +183,6 @@ func Transaction(ctx *cli.Context) error {
 		if chainConfig.IsShanghai(new(big.Int), 0, 0) && tx.To() == nil && len(tx.Data()) > int(chainConfig.MaxInitCodeSize()) {
 			r.Error = errors.New("max initcode size exceeded")
 		}
-		if chainConfig.IsOsaka(new(big.Int), 0) && tx.Gas() > params.MaxTxGas {
-			r.Error = errors.New("gas limit exceeds maximum")
```
amsanghi, line 187: Do you think it might be better to just skip when chain is arbitrum like i did here https://github.com/OffchainLabs/go-ethereum/pull/522/files ?

I dont have a strong preference either way, but wanted to get your opinion.

eljobe: Hate to reopen this can of worms.
But, I kind of agree that it would be better to have the code in here.
That way, the nitro fork can still be run as a proper L1 node (when != arbitrum)
Maybe we could rename the `params` variable to something like:
`params.MaxTxGasRenamedForNitroMerges` (or something more concise, but equally clear.

That way, if the upstream code starts using `MaxTxGas` elsewhere, we'll still notice it, but we can also keep the functionality when arbitrum is disabled?

PR #526, opened 2025-08-29
Enable arbos 50 support

eljobe:
This makes it so that, once the arbos 50 upgrade is applied on chain, the nitro node will still be happy to run it.

Pulled in by: OffchainLabs/nitro/pull/3547


joshuacolvin0 (approved):
LGTM

MishkaRogachev (approved):

Issue #514, opened 2025-08-26
Update dependency on github.com/gammazero/deque v0.2.1


astudnev:
github.com/gammazero/deque v0.2.1

is outdated, please update to 1.X

eljobe:
Why?

astudnev:
It is outdated and new libraries that we use require version 1.X

PR #530, opened 2025-09-03
Update the gammazero/deque dependency

eljobe:
Fixes: NIT-3760
Fixes: #514


Tristan-Wilson (approved):
This is a dependency that we brought in for our metrics patch so it makes sense that it hadn't been updated by upstream and we need to update it.

PR #532, opened 2025-09-03
Multigas: Fix dimension of warm storage read

MishkaRogachev:


core/vm/operations_verkle.go
```diff
@@ -81,33 +80,33 @@ func makeCallVariantGasEIP4762(oldCalculator gasFunc, withTransferCosts bool) ga
 			if wantedValueTransferWitnessGas > contract.Gas {
 				return multigas.StorageAccessGas(wantedValueTransferWitnessGas), nil
 			}
-			witnessGas = wantedValueTransferWitnessGas
+			witnessGas = multigas.StorageAccessGas(wantedValueTransferWitnessGas)
 		} else if isPrecompile || isSystemContract {
-			witnessGas = params.WarmStorageReadCostEIP2929
+			witnessGas = multigas.ComputationGas(params.WarmStorageReadCostEIP2929)
```
MishkaRogachev, line 85: clarify: whether access to precompile and system contract is also computation

PR #533, opened 2025-09-04
ethapi: Update eth_config implementation for Arbitrum chains

eljobe:
There are a couple of important changes:
   1. Arbitrum chains don't support blob posting, so, BlobSchedule is not important.
	 2. The Timestamp method needed to be updated to handle the fact that Arbitrum performs upgrades based on block number, and not timestamp.

Part of: NIT-3679
Pulled in by: https://github.com/OffchainLabs/nitro/pull/3571


joshuacolvin0 (dismissed):
LGTM

PR #537, opened 2025-09-08
core: Add intrinsic gas to GasCharging hook

eljobe:
This GasCharging hook can now deduct the intrinsic gas from the per-transaction gas limit before capping the remaining gas available to the transaction. This is important because intrinsic gas is already deducted before entering the GasCharging hook.

Pulled in by: https://github.com/OffchainLabs/nitro/pull/3589
Part of: NIT-3787


core/state_transition.go
```diff
@@ -602,13 +602,13 @@ func (st *stateTransition) execute() (*ExecutionResult, error) {
 		usedMultiGas     = multigas.ZeroGas()
 	)
 
-	// Check clauses 4-5, subtract intrinsic gas if everything is correct
-	gas, err := IntrinsicGas(msg.Data, msg.AccessList, msg.SetCodeAuthorizations, contractCreation, rules.IsHomestead, rules.IsIstanbul, rules.IsShanghai)
```
hkalodner, line 606: I don't feel at all strongly, but I'd lean against the variable rename in the name of minimizing diff, but defer to others

tsahee: I actually like this rename. It's more diff-lines, but very easy to see you're not doing something weird

PR #539, opened 2025-09-10
Hide multi-gas behind feature flag

gligneul:
Ensure we don't expose a wrong multi-gas before finishing all instrumentation milestones.

Close NIT-3838

### DIFF
--- a/core/vm/interpreter.go
+++ b/core/vm/interpreter.go
@@ -35,6 +35,8 @@ type Config struct {
 
 	StatelessSelfValidation bool // Generate execution witnesses and self-check against them (testing purpose)
 	EnableWitnessStats      bool // Whether trie access statistics collection is enabled
+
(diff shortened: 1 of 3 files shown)


core/vm/interpreter.go
```diff
@@ -35,6 +35,8 @@ type Config struct {
 
 	StatelessSelfValidation bool // Generate execution witnesses and self-check against them (testing purpose)
 	EnableWitnessStats      bool // Whether trie access statistics collection is enabled
+
+	ExposeMultiGas bool // Arbitrum: Expose multi-gas used in transaction receipts
```
eljobe, line 39: Is there somewhere to ensure that it defaults to `false`? Ideally, somewhere that we can later make it default to `true` in some future nitro node release.

eljobe (approved):

eljobe (approved):

MishkaRogachev (approved):

PR #541, opened 2025-09-15
Multigas: Account multi gas callcode delegatecall staticcall

MishkaRogachev:
Part of NIT-3841
Nitro PR:  https://github.com/OffchainLabs/nitro/pull/3635

Update call functions to also return the usedMultiGas:
- [x] evm.CallCode
- [x] evm.DelegateCall
- [x] evm.StaticCall

### DIFF
--- a/core/vm/evm.go
+++ b/core/vm/evm.go
@@ -265,7 +265,6 @@ func (evm *EVM) Call(caller common.Address, addr common.Address, input []byte, g
 	}
 	snapshot := evm.StateDB.Snapshot()
 	p, isPrecompile := evm.precompile(addr)
-	startingGas := gas
 	usedMultiGas = multigas.ZeroGas()
 
 	if !evm.StateDB.Exist(addr) {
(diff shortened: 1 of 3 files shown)


core/vm/evm.go
```diff
@@ -328,7 +329,11 @@ func (evm *EVM) Call(caller common.Address, addr common.Address, input []byte, g
 				evm.Config.Tracer.OnGasChange(gas, 0, tracing.GasChangeCallFailedExecution)
 			}
 
-			usedMultiGas.SaturatingIncrementInto(multigas.ResourceKindComputation, gas)
+			// Attribute all leftover gas to computation
+			if startingGas > usedMultiGas.SingleGas() {
+				leftOverGas := startingGas - usedMultiGas.SingleGas()
+				usedMultiGas.SaturatingIncrementInto(multigas.ResourceKindComputation, leftOverGas)
+			}
```
gligneul, line 336: I didn't get this change. The existing gas variable should be left over gas, isn't it?

core/vm/evm.go
```diff
@@ -377,6 +384,7 @@ func (evm *EVM) CallCode(caller common.Address, addr common.Address, input []byt
 			Evm:               evm,
 		}
 		ret, gas, err = RunPrecompiledContract(p, input, gas, evm.Config.Tracer, info)
+		usedMultiGas.SaturatingIncrementInto(multigas.ResourceKindComputation, gas)
```
gligneul, line 387: Please add a comment that this should be updated on NIT-3557

core/vm/evm.go
MishkaRogachev: Initially, the test results did not match, so I left this form as more informative. I returned usedMultiGas.SaturatingIncrementInto(multigas.ResourceKindComputation, gas)` back.

gligneul (approved):

PR #542, opened 2025-09-15
core: Instrument WasmAccountTouchCost with multigas

MishkaRogachev:
Part of NIT-3796
Nitro PR: https://github.com/OffchainLabs/nitro/pull/3645

- Return multigas from `WasmAccountTouchCost`


gligneul (changes requested):
See comments

PR #638, opened 2026-03-16
consensus: Add CalcBlobFeeWithConfig to eip4844 package

bragaigor:
Adds a new helper function CalcBlobFeeWithConfig that calculates the blob base fee from a params.BlobConfig directly, without requiring a full params.ChainConfig.

The existing CalcBlobFee function requires a `*params.ChainConfig` and a `*types.Header`, then internally resolves the active blob schedule via `latestBlobConfig(config, header.Time)`. This works well when the caller has a complete chain config, but Nitro's ParentChain component now fetches the blob schedule dynamically from the parent chain's `eth_config` RPC (EIP-7910), yielding a standalone `*params.BlobConfig` rather than a full chain config.

Without this helper, the caller would need to either:

- Reconstruct a full `ChainConfig` just to call `CalcBlobFee` (wasteful and error-prone)
- Duplicate the `blobBaseFee` calculation logic outside of geth (breaks encapsulation)

pulled in by https://github.com/OffchainLabs/nitro/pull/4511
part of [NIT-4354](https://linear.app/offchain-labs/issue/NIT-4354/use-eth-config-rpc-method-to-get-bpo-parameters-of-parent-chain)


gligneul (approved):
Looks good. Just a minor question.

gligneul (approved):

PR #642, opened 2026-03-24
arbitrum: Add filtered transactions report types

MishkaRogachev:
Part of NIT-4642
Pulled in by https://github.com/OffchainLabs/nitro/pull/4557

- [x] TouchAddress accepts FilteredAddressRecord (address + reason) instead of common.Address. 
- [x] IsFiltered returns (bool, []FilteredAddressRecord) to propagate which addresses were filtered and why. 
- [x] Adds arbitrum/filter package with report types (FilterReasonType, FilteredAddressRecord).

### DIFF
--- a/arbitrum/filter/filter_report.go
+++ b/arbitrum/filter/filter_report.go
@@ -44,14 +44,14 @@ type FilteredAddressRecord struct {
 }
 
 type FilteredTxReport struct {
-	ID                     string                  `json:"id"`
-	TxHash                 common.Hash             `json:"txHash"`
-	TxRLP                  hexutil.Bytes           `json:"txRLP"`
-	FilteredAddresses      []FilteredAddressRecord `json:"filteredAddresses"`
-	BlockNumber            uint64                  `json:"blockNumber"`
-	ParentBlockHash        common.Hash             `json:"parentBlockHash"`
-	PositionInBlock        uint64                  `json:"positionInBlock"`
-	FilteredAt             time.Time               `json:"filteredAt"`
-	IsDelayed              bool                    `json:"isDelayed"`
-	DelayedInboxRequestId  *common.Hash            `json:"delayedInboxRequestId,omitempty"`
+	ID                    string                  `json:"id"`
+	TxHash                common.Hash             `json:"txHash"`
+	TxRLP                 hexutil.Bytes           `json:"txRLP"`
+	FilteredAddresses     []FilteredAddressRecord `json:"filteredAddresses"`
+	BlockNumber           uint64                  `json:"blockNumber"`
+	ParentBlockHash       common.Hash             `json:"parentBlockHash"`
+	PositionInBlock       uint64                  `json:"positionInBlock"`
+	FilteredAt            time.Time               `json:"filteredAt"`
+	IsDelayed             bool                    `json:"isDelayed"`
+	DelayedInboxRequestId *common.Hash            `json:"delayedInboxRequestId,omitempty"`
 }
--- a/arbitrum/filter/filter_report.go
+++ b/arbitrum/filter/filter_report.go
@@ -36,13 +36,11 @@ type EventRuleMatch struct {
 	RawLog            *RawLog `json:"rawLog,omitempty"`
 }
 
-// lint:require-exhaustive-initialization
 type FilterReason struct {
 	Reason FilterReasonType `json:"reason"`
 	*EventRuleMatch
 }
 
-// lint:require-exhaustive-initialization
 type FilteredAddressRecord struct {
 	Address     common.Address `json:"address"`
 	FilterSetId string         `json:"filterSetId"`
--- a/arbitrum/filter/filter_report.go
+++ b/arbitrum/filter/filter_report.go
@@ -36,13 +36,14 @@ type EventRuleMatch struct {
 	RawLog            *RawLog `json:"rawLog,omitempty"`
 }
 
+// lint:require-exhaustive-initialization
 type FilterReason struct {
 	Reason FilterReasonType `json:"reason"`
 	*EventRuleMatch
 }
 
+// lint:require-exhaustive-initialization
 type FilteredAddressRecord struct {
-	Address     common.Address `json:"address"`
-	FilterSetID string         `json:"filterSetId"`
+	Address common.Address `json:"address"`
 	FilterReason
 }
--- a/arbitrum/filter/filter_report.go
+++ b/arbitrum/filter/filter_report.go
@@ -38,17 +38,20 @@ type FilterReason struct {
 }
 
 type FilteredAddressRecord struct {
-	Address common.Address `json:"address"`
+	Address     common.Address `json:"address"`
+	FilterSetId string         `json:"filterSetId"`
 	FilterReason
 }
 
 type FilteredTxReport struct {
-	TxHash            common.Hash             `json:"txHash"`
-	TxRLP             hexutil.Bytes           `json:"txRLP"`
-	FilteredAddresses []FilteredAddressRecord `json:"filteredAddresses"`
-	BlockNumber       uint64                  `json:"blockNumber"`
-	ParentBlockHash   common.Hash             `json:"parentBlockHash"`
-	PositionInBlock   uint64                  `json:"positionInBlock"`
-	FilteredAt        time.Time               `json:"filteredAt"`
-	IsDelayed         bool                    `json:"isDelayed"`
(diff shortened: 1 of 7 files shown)


core/state/statedb_arbitrum.go
```diff
@@ -312,11 +313,12 @@ var ErrArbTxFilter error = errors.New("internal error")
...
 	// TouchAddress records an address access and checks if it should be filtered.
-	TouchAddress(addr common.Address)
+	TouchAddress(record filter.FilteredAddressRecord)
 
-	// IsFiltered returns whether any touched address was filtered.
+	// IsFiltered returns whether any touched address was filtered and the
+	// list of filtered address records collected during the transaction.
 	// Blocks until all pending checks complete (implementation-specific).
-	IsFiltered() bool
+	IsFiltered() (bool, []filter.FilteredAddressRecord)
```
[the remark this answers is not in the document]
diegoximenes: I am OK with both approaches.
Today, for regular and delayed sequencing, every time IsFiltered is called, and returns true, then FilteredRecords will need to be called to build the report.
This is why I suggested to return both values in the same call.
But this will likely not be true for eth_call and eth_estimategas filtering, so it makes sense to have those two functions now.

core/state/statedb.go
```diff
@@ -221,17 +222,17 @@ func (s *StateDB) SetAddressChecker(checker AddressChecker) {
 	s.arbExtraData.addressChecker = checker
 }
 
-func (s *StateDB) TouchAddress(addr common.Address) {
+func (s *StateDB) TouchAddress(record filter.FilteredAddressRecord) {
```
diegoximenes:
```suggestion
func (s *StateDB) TouchAddress(record *filter.FilteredAddressRecord) {
```

arbitrum/filter/filter_report.go
```diff
@@ -0,0 +1,57 @@
...
+	RawLog            *RawLog          `json:"rawLog,omitempty"`
+}
+
+type FilteredAddressRecord struct {
+	Address     common.Address `json:"address"`
+	FilterSetId string         `json:"filterSetId"`
+	FilterReason
+}
+
+type FilteredTxReport struct {
```
diegoximenes: Lets try to use this lint rule, for this and other structs here.
I want to avoid the issue in the related PR, in which a filter.ReasonEventRule without filling the other fields.
```suggestion
// lint:require-exhaustive-initialization
type FilteredTxReport struct {
```

arbitrum/filter/filter_report.go
```diff
@@ -0,0 +1,57 @@
...
+	Address common.Address `json:"address"`
+	Topics  []common.Hash  `json:"topics"`
+	Data    hexutil.Bytes  `json:"data"`
+}
+
+type FilterReason struct {
+	Reason            FilterReasonType `json:"reason"`
+	MatchedEvent      string           `json:"matchedEvent,omitempty"`
+	MatchedTopicIndex int              `json:"matchedTopicIndex,omitempty"`
+	RawLog            *RawLog          `json:"rawLog,omitempty"`
```
diegoximenes: Lets create another struct containing those fields, something like EventRuleMatch, and use as a pointer here

```suggestion
	*EventRuleMatch
```

arbitrum/filter/filter_report.go
```diff
@@ -0,0 +1,57 @@
...
+	ID                    string                  `json:"id"`
+	TxHash                common.Hash             `json:"txHash"`
+	TxRLP                 hexutil.Bytes           `json:"txRLP"`
+	FilteredAddresses     []FilteredAddressRecord `json:"filteredAddresses"`
+	BlockNumber           uint64                  `json:"blockNumber"`
+	ParentBlockHash       common.Hash             `json:"parentBlockHash"`
+	PositionInBlock       uint64                  `json:"positionInBlock"`
+	FilteredAt            time.Time               `json:"filteredAt"`
+	IsDelayed             bool                    `json:"isDelayed"`
+	DelayedInboxRequestId *common.Hash            `json:"delayedInboxRequestId,omitempty"`
```
diegoximenes: Lets create another struct containing DelayedInboxRequestId, something like DelayedReportData, and use as a pointer here

diegoximenes (changes requested):
Lets create some tests to check that FilteredTxReport is being properly JSON marshalled.
It should cover isDelayed being true and false, Reason being ReasonEventRule and other reason type

arbitrum/filter/filter_report.go
diegoximenes: This struct should be in Nitro.

arbitrum/filter/filter_report.go
```diff
@@ -0,0 +1,48 @@
+package filter
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+type FilterReasonType string
+
+const (
```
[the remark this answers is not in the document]
MishkaRogachev: moved to nitro

arbitrum/filter/filter_report.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type FilterReason struct {
+	Reason FilterReasonType `json:"reason"`
+	*EventRuleMatch
+}
+
+type FilteredAddressRecord struct {
+	Address     common.Address `json:"address"`
+	FilterSetId string         `json:"filterSetId"`
```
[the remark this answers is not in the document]
MishkaRogachev: specs use "FilterSetId"

arbitrum/filter/filter_report.go
MishkaRogachev: moved

arbitrum/filter/filter_report.go
MishkaRogachev: fixed

arbitrum/filter/filter_report.go
MishkaRogachev: fixed

arbitrum/filter/filter_report.go
MishkaRogachev: fixed & moved to nitro

core/state/statedb.go
MishkaRogachev: fixed

arbitrum/filter/filter_report.go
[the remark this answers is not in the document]
diegoximenes: I agree with copilot.
filterSetId is the convention for JSON, which is correct here.
FilterSetID is the convention for golang.

diegoximenes (changes requested):

arbitrum/filter/filter_report.go
```diff
@@ -0,0 +1,48 @@
...
+	MatchedTopicIndex int     `json:"matchedTopicIndex,omitempty"`
+	RawLog            *RawLog `json:"rawLog,omitempty"`
+}
+
+type FilterReason struct {
+	Reason FilterReasonType `json:"reason"`
+	*EventRuleMatch
+}
+
+type FilteredAddressRecord struct {
```
diegoximenes, line 46: Also add the `lint:require-exhaustive-initialization` here.

arbitrum/filter/filter_report.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// lint:require-exhaustive-initialization
+type EventRuleMatch struct {
+	MatchedEvent      string  `json:"matchedEvent"`
+	MatchedTopicIndex int     `json:"matchedTopicIndex,omitempty"`
+	RawLog            *RawLog `json:"rawLog,omitempty"`
+}
+
+type FilterReason struct {
```
diegoximenes, line 40: Also add the lint:require-exhaustive-initialization here.

arbitrum/filter/filter_report.go
```diff
@@ -0,0 +1,48 @@
...
+	Reason FilterReasonType `json:"reason"`
+	*EventRuleMatch
+}
+
+type FilteredAddressRecord struct {
+	Address     common.Address `json:"address"`
+	FilterSetID string         `json:"filterSetId"`
```
diegoximenes: Let's remove the FilterSetID for now, and only add it in this [task](https://linear.app/offchain-labs/issue/NIT-4780/add-filtersetid-to-filteredaddressrecord)

diegoximenes (changes requested):

diegoximenes (approved):

tsahee (approved):
LGTM

PR #645, opened 2026-03-31
core: Add Cranelift WasmTarget variants for Stylus overflow fallback

bragaigor:
When a Stylus program overflows the native (Singlepass) coroutine stack, Nitro recompiles it with Cranelift as a fallback. The Cranelift ASM needs to be stored separately from the Singlepass ASM in the wasm store so both remain available — normal execution uses Singlepass, and only the overflow retry path uses Cranelift.

- Add three new `WasmTarget` constants: `TargetArm64Cranelift`, `TargetAmd64Cranelift`, `TargetHostCranelift`
- Add corresponding wasm store key prefixes ({0x00, 'c', 'r/x/h'}) separate from the existing Singlepass prefixes (`{0x00, 'w', 'r/x/h'}`), included in `WasmPrefixesExceptWavm()`
- Add helper functions: `CraneliftTarget()`, `BaseTarget()`, `IsCraneliftTarget()`, `SplitAsmMap()`, `DeduplicateAsmMap()`, `DeleteActivatedAsm()`
- `ActivateWasm`: deduplicate cranelift entries into base targets for the consistency check via `DeduplicateAsmMap`, then eagerly persist cranelift entries to the wasm store under their own keys
- Journal revert (`wasmActivation`) deletes eagerly-written cranelift entries from the wasm store on `RevertToSnapshot`
- ActivatedAsmMap: fall back to cranelift ASM when singlepass is missing for a requested target
- Add unit tests for all new `rawdb` helpers and for the activation/revert path in `StateDB`

pulled in by https://github.com/OffchainLabs/nitro/pull/4538


core/state/statedb_arbitrum.go
```diff
@@ -172,21 +172,27 @@ func NewStylusRootPrefix(dictionary byte) []byte {
 // If the asmMap is added to the newly activated wasms, then wasmActivation is added to the journal so the operation can be reverted and the new entry removed in StateDB.RevertToSnapshot.
 // note: all ActivateWasm calls in given StateDB cycle (cycle reset by statedb commit) requires that the asmMap contain entries for the same targets as the first asmMap passed to ActivateWasm in the cycle. This is assumed in other parts of the code.
 func (s *StateDB) ActivateWasm(moduleHash common.Hash, asmMap map[rawdb.WasmTarget][]byte) error {
+	// Deduplicate: collapse cranelift entries into their base targets so that
+	// the consistency check sees the same target set regardless of whether
+	// individual programs used the cranelift fallback. Singlepass takes
+	// precedence when both exist for the same base target.
+	dedupAsms := rawdb.DeduplicateAsmMap(asmMap)
+
```
bragaigor: I thought it made sense to deduplicate `asmMap` here; however, that means we need to have the `DeduplicateAsmMap` logic in go-ethereum side, which is no small change. We could move this to nitro side along with its logic and tests but we would need to make sure that all calls to `ActivateWasm` are guarded with such call. Personally I would vote to keep it here but open to suggestions

core/rawdb/accessors_state_arbitrum.go
```diff
@@ -59,12 +63,99 @@ func activatedAsmKeyPrefix(target WasmTarget) (WasmPrefix, error) {
...
+	default:
+		return false
+	}
+}
+
+// SplitAsmMap separates a unified asmMap into consensus (non-cranelift) and
+// cranelift entries. This allows callers to pass only consensus targets to
+// ActivateWasm (which has consistency checks) while persisting cranelift
+// entries separately to the wasm store.
+func SplitAsmMap(asmMap map[WasmTarget][]byte) (consensus map[WasmTarget][]byte, cranelift map[WasmTarget][]byte) {
```
tsahee: look

core/state/statedb_arbitrum_test.go
```diff
@@ -49,3 +50,67 @@ func TestRecentWasmsInsertAndCopy(t *testing.T) {
 		t.Fatalf("original: first insert of hash3 should be a miss (must be independent of copy)")
 	}
 }
+
+func TestActivateWasmRevert(t *testing.T) {
```
bragaigor, line 54: lightweight test that now validates the new behavior (cranelift entries stored alongside singlepass). I think it's nice to have a test to make sure activation and revert happens for both.

eljobe (approved):

PR #646, opened 2026-04-01
params: Add DisableOffchainArbOwner chain config parameter

bragaigor:
- Add `DisableOffchainArbOwner` bool field to `ArbitrumChainParams`
- Add `DisableOffchainArbOwner()` accessor method on `*ChainConfig`
- Defaults to `false` (no behavior change for existing chains)

pulled in by https://github.com/OffchainLabs/nitro/pull/4591
part of [NIT-4744](https://linear.app/offchain-labs/issue/NIT-4744/add-a-flag-to-disable-arbowner-unless-in-isexecuteonchain-mode)


bragaigor:
going with a different approach so closing it

PR #649, opened 2026-04-10
rpc: support wrapped errors and code-based Is for jsonError

pmikolajczyk41:
1. `errorMessage` previously cast err directly to `rpc.Error`/`DataError`, which missed errors wrapped with `fmt.Errorf("%w", ...)`. Switch to `errors.As` so the code and data are extracted correctly from wrapped errors.
2. Add `(*jsonError).Is` so that `errors.Is(receivedErr, sentinel)` works on the client side when `receivedErr` is a `jsonError` reconstructed from the wire. Matching is by error code: if the sentinel implements `ErrorCode() int`, the codes are compared. This avoids the need for string-based error reconstruction in callers.

---

pulled in by https://github.com/OffchainLabs/nitro/pull/4629

### DIFF
--- a/rpc/json.go
+++ b/rpc/json.go
@@ -126,12 +126,12 @@ func errorMessage(err error) *jsonrpcMessage {
 		Code:    ErrcodeDefault,
 		Message: err.Error(),
 	}}
-	ec, ok := err.(Error)
-	if ok {
+	var ec Error
+	if errors.As(err, &ec) {
 		msg.Error.Code = ec.ErrorCode()
 	}
-	de, ok := err.(DataError)
-	if ok {
+	var de DataError
+	if errors.As(err, &de) {
 		msg.Error.Data = de.ErrorData()
 	}
 	return msg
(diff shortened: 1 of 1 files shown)


rpc/json.go
```diff
@@ -126,12 +126,12 @@ func errorMessage(err error) *jsonrpcMessage {
 		Code:    ErrcodeDefault,
 		Message: err.Error(),
 	}}
-	ec, ok := err.(Error)
-	if ok {
+	var ec Error
+	if errors.As(err, &ec) {
```
diegoximenes, line 130: It is annoying that the upstream geth doesn't use error.As already here.
But changing our fork like this can hurt us down the line, when updating the fork with upstream geth.
Also, not sure if this will negatively affect native geth APIs.
So avoiding changes like this in geth, if the alternative is not too complicated, is advisable.

An alternative is, the server side, e.g. execution/rpcserver.Server. DigestMessage, checks if the error to be returned has a wrapped RPCError through error.As. If so then it returns a plain RPCError, with the same error code of the wrapped RPCError, and error message equal to the original error.

rpc/json.go
```diff
@@ -158,6 +158,19 @@ func (err *jsonError) ErrorData() interface{} {
 	return err.Data
 }
 
+// Is reports whether target carries the same JSON-RPC error code as err.
+// This allows errors.Is(receivedErr, sentinel) to work on the client side
+// when receivedErr is a jsonError reconstructed from the wire and sentinel
+// is any error that exposes an ErrorCode() int method.
+func (err *jsonError) Is(target error) bool {
```
diegoximenes, line 165: This one can be avoided if the client side, e.g. execution/rpcclient.Client.DigestMessage, always tries to build a RPCError, from the error returned by the server, using the same error code and message provided by the error returned by the server.
RPCError already has a Is func.

diegoximenes (changes requested):

rpc/json.go
pmikolajczyk41: proposed this change to upstream: https://github.com/ethereum/go-ethereum/pull/34886

